Issue #141, opened 2026-07-18
[Bug]: Crowded/wrapped text in advanced compat adapters


jamezrin:
### What happened?

Some items in `<select>` elements are too large and they make other content (at the left) wrap to adjust horizontal length/width.

<img width="400" height="594" alt="Image" src="https://github.com/user-attachments/assets/a2427d81-79f7-4cfd-bd89-53b07f594120" />

### Steps to reproduce

Open compatibility settings in the advanced section

### Where are you running Lurkloot?

Chrome or Chromium extension

### Lurkloot version

1.5.0

### Browser and operating system

Chrome

### Affected streaming platform

Not platform-specific

### Watch mode

Not applicable

### Relevant logs

```shell

```

### Before submitting

- [x] I searched the existing issues for this problem.
- [x] I removed passwords, cookies, tokens, authorization headers, and other credentials from this report.

PR #143, opened 2026-07-18
feat(popup): restructure compatibility settings around resolved values

jamezrin:
Stacked on #142 — base is `fix/compat-select-row-width`, so review that one first. Retarget to `develop` once #142 merges.

## Why

The section was counterintuitive for structural reasons, not cosmetic ones:

1. **The control and its result lived in two places.** You picked "Automatic" in a select; what it *resolved to* appeared in a separate "Effective compatibility" block below.
2. **That block listed five capabilities but only two controls were visible** until you expanded "expert" — it reported on rows you couldn't see.
3. **The profile→component hierarchy was invisible.** A profile bundles a heartbeat + inventory; the expert selects override what the profile chose. Nothing said so.
4. **Five identical selects reading "Automatic"**, none showing what they picked.
5. **Twitch and Kick were interleaved** rather than grouped.
6. **Lifecycle badges appeared everywhere except where you choose** — in the effective list and a floating legend, but not on the dropdown options.
7. **`replacement` metadata was never surfaced**, so a legacy pick never said what supersedes it.

## What changed

- **Resolved value folded into each row** — title, lifecycle badge and raw id sit on the control that produced them. "Automatic" now always states what it actually resolved to. The separate `compatibilityEffectiveTitle` block is gone.
- **Grouped by platform** with the existing `PLATFORMS` mark/colour convention already used in platform settings.
- **Provenance is explicit** — each component says `From profile` or `Overridden`; the section header carries an Automatic/Overridden pill.
- **Components stay read-only until expert mode.** The dangerous knobs remain gated, but their resolved state is always visible — that is what fixes the "reports on rows you can't see" problem.
- **Lifecycle moved into option labels** (`Spade heartbeat v1 · Recommended`), where the tradeoff matters at the moment of choosing. This also drops the raw `(id)` suffix that caused #141 in the first place.
- **Legacy picks name their successor** via the registry's `replacement` field (`Replaced by Spade heartbeat v1`).
- **`SelectControl` extracted** from `SelectSettingRow` so the width cap from #142 lives in one place and the compat rows reuse it.

Accessible names are unchanged: the controls keep the full capability names (`Twitch heartbeat transport`) as `aria-label` while the visible labels shorten to `Heartbeat transport` under the platform heading.

No change to the settings shape, the resolver, or `@lurkloot/core` — this is popup-ui plus copy.

## Testing

`pnpm check` passes — 511 extension tests (+2 net), all workspace typechecks, site build.

`compatibilitySettingsView.test.tsx` updated to the new IA, with new coverage for what the redesign introduces: resolved ids rendered per row, lifecycle carried into option labels, and `From profile` → `Overridden` + replacement hint when a legacy option is picked. The test's translate stub now performs `$1` substitution like `translateFromCatalogs` does, so the placeholder path is actually exercised.

Verified in the real component at the real 400px popup width by temporarily wiring the compat registry into the site demo (reverted before commit — the demo adapter omits it, so the section doesn't otherwise render there). Checked: default state, expert mode, an active override, dark mode via `prefers-color-scheme`, and Arabic RTL. No horizontal overflow in any of them (360px content in a 360px box).

## Please look at

- **The 9 new strings across 10 catalogs are machine-translated.** They pass the repo's parity and not-left-as-English checks, but the non-English wording wants a native review — particularly the German `Einlöse-Verarbeitung` and Russian `Обработка получения наград` for "Claim handling", which are the least idiomatic.
- **Registry `description` fields are still unused.** They are English-only literals in `@lurkloot/core`, so surfacing them would put untranslated text in a localized UI. I used the existing translated description keys instead — worth deciding if those descriptions should move into the catalogs.
- **`45%` select cap** inherited from #142 still applies; with the `(id)` suffix gone from option labels it now rarely engages.

## Summary by CodeRabbit

* **New Features**
  * Redesigned compatibility settings with component-level options and clearer lifecycle information.
  * Added indicators showing whether settings come from a profile or are manually overridden.
  * Added replacement guidance for legacy options and a control to restore automatic settings.
  * Improved selection controls with consistent styling and helpful hover descriptions.

* **Documentation**
  * Updated compatibility terminology and translations across supported languages.

* **Tests**
  * Expanded coverage for platform grouping, lifecycle labels, overrides, replacements, and restoring automatic settings.

### DIFF
--- a/packages/extension/tests/compatibilitySettingsView.test.tsx
+++ b/packages/extension/tests/compatibilitySettingsView.test.tsx
@@ -14,7 +14,15 @@ const labels: Record<string, string> = {
   compatibilityTwitchProfileDescription: "Twitch profile description",
   compatibilityKickProfileTitle: "Kick compatibility profile",
   compatibilityKickProfileDescription: "Kick profile description",
-  compatibilityEffectiveTitle: "Effective compatibility",
+  compatibilitySectionTitle: "Compatibility",
+  compatibilitySectionDescription: "Compatibility description",
+  compatibilityComponentProfile: "Profile",
+  compatibilityComponentHeartbeat: "Heartbeat transport",
+  compatibilityComponentInventory: "Inventory query",
+  compatibilityComponentClaim: "Claim handling",
+  compatibilityFromProfile: "From profile",
+  compatibilityOverridden: "Overridden",
+  compatibilityReplacedBy: "Replaced by $1",
   compatibilityExpertShow: "Show expert compatibility controls",
   compatibilityExpertHide: "Hide expert compatibility controls",
   compatibilityTwitchHeartbeatTitle: "Twitch heartbeat transport",
@@ -59,7 +67,14 @@ function mount(onChange = vi.fn()) {
       onChange(patch);
       setSettings((current) => applySettingsPatch(current, patch));
     };
-    return <I18nContext.Provider value={{ t: (key) => labels[key] ?? key, dir: "ltr", locale: "en" }}>
+    // Mirrors translateFromCatalogs so $1 placeholders resolve like they do in
+    // the real popup.
+    const translate = (key: string, substitutions?: string | string[]) => {
+      const template = labels[key] ?? key;
+      const values = Array.isArray(substitutions) ? substitutions : substitutions == null ? [] : [substitutions];
+      return values.reduce((text, value, index) => text.replaceAll(`$${index + 1}`, value), template);
+    };
+    return <I18nContext.Provider value={{ t: translate, dir: "ltr", locale: "en" }}>
       <CompatibilitySettings
         settings={settings.compatibility}
         registry={COMPATIBILITY_REGISTRY}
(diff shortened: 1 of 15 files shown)


jamezrin:
Pushed 1195c9f — removed the expert disclosure.

You were right that it didn't make sense. Once each row carried its own resolved value, the toggle stopped revealing anything: it kept a disclosure chevron but only swapped rows between read-only and editable, so it read as a collapsed container that showed nothing when you opened it. That was a regression I introduced in the restructure — before it at least revealed genuinely hidden rows.

Every capability is now overridable directly, no disclosure step. The gate wasn't much of a safeguard regardless: the override warning and "Restore automatic compatibility" already cover the risk and both stay.

Also removes the now-unused `compatibilityExpertShow` / `compatibilityExpertHide` keys from all ten catalogs, and the expert state from `SettingsView`.

`pnpm check` still green — 511 tests. The view test no longer needs a disclosure click to reach the component selects, which is a small sign the flow got simpler.

PR #236, opened 2026-07-24
feat(settings): decouple campaign farming eligibility from the drops list filter

jamezrin:
## Summary

Two changes, one idea: **campaign filters now decide what gets farmed, not just what's shown — split into two clean axes, with a safety rule that you can never farm a campaign you can't see.**

In the shipped version the campaign filters only affect the Drops list; the scheduler ignores them, so you can hide a campaign and it keeps being farmed invisibly. This replaces that with:

- **Farming eligibility** — two toggles that decide what the scheduler farms: *Farm campaigns without a linked account* and *Farm campaigns that require a subscription*. Both default on.
- **Drops list view** — six chips that decide what the list shows: `upcoming`, `expired`, `finished`, `excluded`, `notLinked`, `subscription`. Pure display.

**Nothing changes under default settings.**

## The safety invariant

**Anything that will be farmed is always visible in the Drops list** — you should never be farming a campaign you can't see. This is enforced in `isCampaignVisible`, not just the UI: a not-linked or subscription campaign can be hidden *only* when you're not farming that class (`farming-on OR show-flag-on`). In the settings UI this surfaces as a **lock**: when a farming toggle is on, its matching view chip is forced on, disabled, marked with a lock glyph + tooltip, and — crucially — its stored preference is left untouched, so if you later stop farming that class your earlier show/hide choice returns.

The four lifecycle chips (`upcoming`/`expired`/`finished`/`excluded`) need no lock: none of those states is ever farmable, so hiding them can't hide anything farmed. A binding test ties `isCampaignVisible` to the engine's farming predicate, so a future change that could hide a farmable campaign fails CI.

## Why this shape

The two axes sit on the correct side of the engine/extension boundary:
- `EngineSettings.farmingEligibility` — behaviour the scheduler enforces (`isEligible` → `campaignPassesFarmingEligibility`), honoured by the **CLI** as a validated knob.
- `ExtensionSettings.dropsListFilter` — a popup view preference the engine never reads and the **CLI rejects as extension-only** (a headless run has no list to filter).

## Exclusion is unchanged

Excluded campaigns were never farmed and still aren't (`excludedCampaignIds`, enforced by `isEligible`). The `excluded` chip only decides whether they still appear in the list. A claimable reward always keeps a campaign visible regardless of any chip.

## Migration

Schema v2 (unshipped) maps all six old `campaignVisibility` keys into `dropsListFilter`, values verbatim — so your existing display preferences (including hidden not-linked/subscription campaigns) are **preserved**. It never populates `farmingEligibility`, which defaults on, so **no profile's farming changes on upgrade**. If you'd hidden not-linked campaigns and they're now being farmed by default, the invariant keeps them visible (chip locked-on) while retaining your hide preference for when farming is off. The interim `campaignFilters` name from this branch's earlier commits never shipped and is dropped.

## Accessibility & polish

The view chips are a `role="group"` with an accessible name; each chip carries `aria-pressed`, and locked chips add `aria-disabled` + a tooltip + a lock glyph. Active chips use `--accent-contrast`, fixing white-on-green illegibility on the Kick accent.

## Testing

`pnpm verify` from the worktree root, exit 0:
- workspace typechecks clean; `@lurkloot/cli` 125, `@lurkloot/extension` 847, `@lurkloot/site` 3 — all passing
- Chromium MV3 and Firefox MV2 builds succeed, 11 locale catalogs bundled

Notable coverage: the **invariant** (farmed campaign visible with every display flag off, including farming-on overriding a hidden chip); the **decoupling** (a subscription campaign you opted out of farming is still visible); the **lock** cannot mutate the stored flag; the v2 migration maps all six keys with values intact and never touches farming (guard test: legacy `notLinked: false` does not set `farmUnlinkedCampaigns: false`); CLI accepts/validates `farmingEligibility` and rejects `dropsListFilter`; an end-to-end storage test through the real `loadSettings`.

## Manual verification still required before merge

The migration has never run against real persisted data. Load the new build against a profile with settings from a released version, open **Settings → Drops**, and confirm your saved toggle/chip states carried over (not defaults). If anything comes back defaulted, the migration lost data and this must not merge.

## Screenshots

Popup settings UI changed substantially — before/after of Settings → Drops expected, ideally including dark theme and a farming-locked chip.

## Follow-up (not in this PR)

The Drops **empty state** — "why is nothing being farmed / shown?" — remains the runtime unifier across the farming controls (categories, eligibility, exclusions, priority mode), speced separately and unblocked by this decoupling.


jamezrin:
Thanks — went through all six. Addressed in `1e065fa`.

**🟠 Migration compatibility claim (architecture.md) — valid, and it was a real bug, not just wording.** In every shipped release `campaignVisibility` is display-only, so its `notLinked`/`subscription` values are pure list-display preferences. The migration was carrying them into the new `farmingEligibility` gates, which would have silently *stopped farming* unlinked/subscription campaigns for anyone who had merely hidden them from their list. Fixed: the migration now maps **only** the four lifecycle keys (`upcoming`/`expired`/`finished`/`excluded`) into `dropsListFilter` and drops `notLinked`/`subscription`, so `farmingEligibility` is never populated by migration and normalization defaults both flags to on. Farming behaviour is now genuinely unchanged for every profile, and the doc says so accurately.

This is backed by a new governing invariant (below), which also means the previously-hidden unlinked/subscription campaigns reappearing in the list is correct, not a regression: they're farmed, so they must be visible.

**🟡 Eligibility-only lifecycle states (campaignFilters.ts) — valid, fixed.** `isCampaignExpired`/`isCampaignFinished` now also honour `eligibility === "expired"`/`"completed"`, and a new `isCampaignUpcoming` covers `status === "upcoming"` OR `eligibility === "upcoming"`, used in both `campaignFilterCategories` and `isCampaignVisible`. Display now matches the engine's lifecycle view. Safe against the invariant since all these states are non-farmable.

**🟡 Claimable-reward exception (spec) — valid, documented.** The spec's Migration/Invariant section now states that a campaign with a claimable reward stays visible even when its lifecycle state is filtered out.

**New invariant encoded.** Anything the engine would farm is always visible in the Drops list — a binding test ties `isCampaignVisible` (all display flags off) to the farming predicate, so a future change that could hide a farmable campaign fails CI.

**🟡 Plan-doc nits (test/ vs tests/ path, superseded `ExtensionSettings["campaignFilters"]` type ref, status.ts contradiction):** these are in the R1 implementation-plan doc, which Revision 2 in the spec supersedes and which no longer matches the shipped code. Left as historical process artifacts rather than back-edited; the shipped code uses the correct paths and the engine-owned contract.

PR #238, opened 2026-07-24
fix(controller): keep probed auth health when a tick rolls back

jamezrin:
## Summary

Fixes Twitch (and Kick) getting stuck on **"Checking your signed-in session…"** in the popup for seemingly no reason.

That message shows whenever `authHealth.status === "checking"`, which is meant to be transient. The controller's `tick()` coupled its resolution to the success of the *entire* scheduler tick:

- It probed auth health into an in-memory `schedulerState` and only persisted it after `runSchedulerTick` + notifications + ad-focus + tabless reconcile all succeeded.
- If **anything** downstream threw, the `catch` re-persisted the **pre-tick** state, discarding the fresh probe and reverting auth health to its prior value — `checking` after a credential-cookie change (`invalidateAuthHealth`) or a fresh service-worker start.
- Separately, `probeAuthHealth` awaited `deps.checkCredentialAvailability` (a live cookie read) **unguarded**, so a transient failure there threw straight out of the tick and triggered the same rollback — right at the auth check.

## Fix

- `probeAuthHealth` now always resolves to a terminal status: a throwing credential read or adapter probe maps to `unavailable` instead of propagating into the tick.
- The tick remembers each platform's probed health and **re-applies it on the rollback path**, so a scheduler-body failure keeps the auth status instead of reverting to `checking`.

## Testing

- Added a regression test in `backgroundController.test.ts` reproducing the stuck state (credential probe throws → status was pinned on `checking`); fails before, passes after.
- `pnpm test` — extension 809, cli 117, site 3, all passing.
- `pnpm typecheck` — clean across all packages.

Note: the credential-probe path is covered by the new test. The rollback-preservation branch is a defensive change verified by reasoning + typecheck (couldn't cheaply force one of the heavily-guarded downstream tick steps to throw in a unit test).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication health handling when credential checks or platform probes fail.
  * Prevented authentication status from remaining stuck in “checking” after scheduler errors.
  * Ensured failed checks report a clear “unavailable” status and persist correctly after recovery.

* **Tests**
  * Added coverage for credential probe failures and the resulting authentication status transition.


jamezrin:
Addressed the review: added a test in `backgroundController.test.ts` (`preserves a healthy auth-health probe when a later scheduler step throws`) where `checkAuthHealth` resolves healthy but a later step (`reconcileTablessWatchers` creating the tabless watcher) throws. It asserts the rollback preserves and persists the resolved `healthy` auth-health instead of reverting to `checking`, and that the tick still rolled back (interruption emitted). Verified it fails without the preservation fix. Existing credential-probe failure coverage is unchanged. (8b979a3)

Issue #247, opened 2026-07-25
feat(scheduler): honor a manual watch tab close instead of reopening it


jamezrin:
Follow-up to #242 / #243, from the same user report.

The user repeatedly closes the Kick watch tab and the extension keeps bringing it back: *"after I close the Kick tab ... it returns me to the newly auto-opened Kick tab"*. They resort to disabling the extension from a second browser window.

## Current behavior (working as designed)

`controller.ts:558` `handleTabRemoved` deliberately does **not** run the scheduler from `tabs.onRemoved` — it clears ownership state and emits *"Managed watch tab was closed; recovery deferred to the next scheduler cycle"* (added in #193 / 04413e0, which fixed a genuine immediate-reopen loop).

But recovery still happens on the next ordinary alarm, roughly a minute later. From the user's side there is no way to say "stop" by closing the tab — the only exit is disabling the extension entirely. Closing a window the software owns is the most natural "stop" gesture a user has, and we currently override it.

## Proposal (reviewer's call — this is a user-facing behavior change)

Treat a manual close of a *managed* watch tab as an intentional pause signal rather than a transient failure:

- On manual close of a managed watch tab, pause farming for that platform instead of scheduling recovery.
- Surface it clearly in the popup — a visible paused state with a one-click resume, so the pause is never silent or mysterious.
- Leave the user's `enabled` / `running` settings intact; this is a pause, not a disable.
- Do not apply this to tabs the extension does not own, or to page-context/fallback tabs (#193 rules still hold).

Alternatives worth weighing in review, since this changes expectations for existing users:
1. Pause only after N closes within a short window, so a single accidental close still recovers.
2. Make it a setting (default on) rather than unconditional.

Start from the simplest version that respects the gesture, and flag the trade-off in the PR rather than silently picking one.

## Scope

- `packages/core/src/background/controller.ts` (`handleTabRemoved`), scheduler state, and the popup surface for the paused state.
- Localized copy for any new user-visible string, across every catalog in `packages/locales/messages/`.
- Deterministic unit tests with mocked browser APIs.

PR #248, opened 2026-07-25
fix(tabs): bound playback priming so it stops stealing focus

### LINKED ISSUE #246: fix(tabs): playback priming steals foreground focus on a schedule
Follow-up to #242. That issue fixed the *ad-focus* path stealing focus; `primeTabPlayback` is a second, independent mechanism that does the same thing and is still unfixed.

Same user report (Kick farming, RU, on 1.8.1): after they close the Kick tab, a new one auto-opens and *"it starts dodging my movements to disable the extension"* — they cannot reach the extensions menu. #242's ad false positive explains part of this, but not the burst that happens immediately on tab creation, with no ad involved.

## Current behavior

`packages/core/src/core/tabs.ts:167` — `primeTabPlayback` foreground-activates the watch tab, waits `PLAYBACK_PRIME_RESTORE_DELAY_MS` (1500ms), then restores the previously active tab:

```ts
await browserApi.tabs.update(tabId, { active: true });
await wait(playbackPrimeRestoreDelayMs());
if (previousActiveId != null && previousActiveId !== tabId) {
  await browserApi.tabs.update(previousActiveId, { active: true });
}
```

It runs:
- on **every** managed watch tab creation when `keepVideosUnmuted` is on — the default (`background.ts:105`, `keepFarmingVideosUnmuted !== false`);
- on tab **reuse** whenever `shouldPrimePlayback` returns true (`tabs.ts:154`), which includes `playback.playingVideoCount === 0` and `videoCount === 0`.

## Why it is a problem

In a muted background watch tab where the browser blocks or stalls playback, `playingVideoCount` stays 0, so `shouldPrimePlayback` is true on **every scheduler tick** — producing an indefinite 1.5-second foreground flicker every cycle, with no ad and no end condition. During that window the user's click on the toolbar or the extensions menu is stolen.

Restoring focus afterwards does not make this acceptable: the steal itself dismisses open browser UI (the extensions popup closes the moment focus moves), and it repeats forever.

## Scope

- Do not re-prime indefinitely. Bound retries per tab/session — after N failed priming attempts, stop trying and surface a diagnostic instead of flickering forever.
- Consider back-off between attempts rather than priming on every tick.
- Consider not priming at all when the video is playing-but-muted, which already satisfies farming (watch time is credited while muted — see the comment at `tabs.ts:160`).
- Keep core browser-free; the `*WithBrowser` injection seam and `coreBoundary.test.ts` must keep passing.

Priming exists to coax a deferred player into loading. That goal does not require an unbounded loop.

jamezrin:
Closes #246

## Summary

`primeTabPlayback` could foreground-activate the managed watch tab forever. It is now bounded and backed off, in `packages/core/src/core/tabs.ts`:

- **Per-tab priming budget.** A module-scope `Map<number, PlaybackPrimeState>` tracks attempts and the last attempt timestamp. New `maybePrimeTabPlayback` gates every call site; `primeTabPlayback` itself is unchanged.
- **Cap.** `PLAYBACK_PRIME_MAX_ATTEMPTS` (3), declared beside the existing `PLAYBACK_PRIME_RESTORE_DELAY_MS`. On the tick after the cap is reached the tab is marked exhausted, a `warn` diagnostic is emitted once, and the tab is never activated again for that session.
- **Back-off.** `PLAYBACK_PRIME_BACKOFF_MS` (5 minutes) between attempts, so priming no longer happens on every scheduler tick.
- **Reset points.** `resetPlaybackPriming(tabId?)` is exported and called when a tick finds playback healthy (`shouldPrimePlayback` false), when a managed/reused tab is retargeted at a different channel URL, when a stale managed tab is removed, and when a new tab is created (so a recycled tab id starts fresh). A deferred player on a fresh tab is still coaxed into loading exactly as before.

Core stays browser-free: no browser globals or WXT imports were added, only `Date.now()` and the existing `*WithBrowser` seam. `coreBoundary.test.ts` passes. `controller.ts` and `handleTabRemoved` were **not** touched (#247 owns those).

## Root cause

`shouldPrimePlayback` returns true whenever `playback.playingVideoCount === 0`. In a muted background tab whose playback the browser blocks or stalls, that condition never clears, and the scheduler calls `openPinnedMutedTab` on every tick with `keepVideosUnmuted` on (the default). Each call ran the full prime: `tabs.update(tabId, { active: true })`, wait 1500ms, restore the previous tab. Nothing counted attempts and nothing spaced them, so the flicker repeated indefinitely.

## User impact

A watch tab whose player will not start now pulls itself to the foreground at most 3 times, spaced 5 minutes apart, instead of every tick forever. The browser toolbar and the extensions menu stay reachable, and the log explains why priming stopped rather than leaving the user with an unexplained flicker.

## Not changed

The "skip priming when the video is playing but muted" idea from the issue is **already** the current behaviour and needed no change: `shouldPrimePlayback` only keys off `videoCount === 0 || playingVideoCount === 0`, never the muted counts, and `tabs.test.ts` already asserts it with "does not re-prime a matching managed tab that is playing but muted". The comment at `tabs.ts:160` documents this. Nothing was touched there.

## Testing

TDD: tests written first and observed failing (32 failures in `tabs.test.ts`, starting from the missing `resetPlaybackPriming` export), then implemented.

New cases in `packages/extension/tests/tabs.test.ts`, all with a mocked browser and `vi.useFakeTimers()` / `vi.setSystemTime` (no real timers, no live calls):

- *stops priming a managed tab whose playback never starts* — the unbounded-repeat regression: `PLAYBACK_PRIME_MAX_ATTEMPTS + 3` ticks with a permanently non-playing video produce exactly `PLAYBACK_PRIME_MAX_ATTEMPTS` foreground activations plus a `warn` diagnostic.
- *backs off instead of priming a managed tab on every tick* — a second tick 1s later does not prime; a tick after the back-off does.
- *allows priming again once playback recovers on a managed tab* — an exhausted budget is cleared by one healthy tick.

`pnpm check`: exit 0 — 6/6 package typechecks Done, 47 extension test files / 868 tests passed, 9 CLI files / 125 tests passed, script tests 17 + 70 passed with 0 failures, Astro site build complete.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved playback startup reliability by capping repeated watch-tab playback priming attempts and adding a retry backoff.
  - Prevented repeated tab activations when playback remains stalled, with automatic recovery once playback becomes healthy again.
  - Reset the priming attempt budget when switching channels or changing the target channel identity.

- **Tests**
  - Added timer-based test coverage for max retry limits, backoff behavior, state reset on recovery, and cap reset across channel changes.

### DIFF
--- a/packages/core/src/core/tabs.ts
+++ b/packages/core/src/core/tabs.ts
@@ -61,9 +61,10 @@ const PLAYBACK_PRIME_RESTORE_DELAY_MS = 1500;
 // become an open-ended loop: a tab whose player the browser permanently blocks
 // keeps reporting playingVideoCount === 0, which would otherwise flicker the tab
 // to the foreground on every scheduler tick and make the browser toolbar and the
-// extensions menu unusable. Attempts are counted per tab, spaced by a back-off,
-// and give up after the cap with a warning. The counter resets as soon as a tick
-// finds playback healthy, so a genuinely deferred player is still coaxed along.
+// extensions menu unusable. Attempts are counted per watch target, spaced by a
+// back-off, and give up after the cap with a warning. The counter resets as soon
+// as a tick finds playback healthy, so a genuinely deferred player is still
+// coaxed along.
 const PLAYBACK_PRIME_MAX_ATTEMPTS = 3;
 const PLAYBACK_PRIME_BACKOFF_MS = 5 * 60_000;
 const PAGE_CONTEXT_RECOVERY_SUCCESSES = 3;
@@ -87,12 +88,11 @@ export async function openPinnedMutedTabWithBrowser(
         if (Object.keys(updateProperties).length > 0) {
           await browserApi.tabs.update(tab.id, updateProperties);
         }
-        // A retargeted tab is a fresh player, so its priming budget starts over.
-        if (tab.url !== channel.url) resetPlaybackPriming(tab.id);
         if (!shouldPrimePlayback(tab, channel.url, session)) {
-          resetPlaybackPriming(tab.id);
+          // Playback is healthy, so the budget has served its purpose.
+          resetPlaybackPriming(channel.platform);
         } else if (tabOptions.keepVideosUnmuted) {
-          await maybePrimeTabPlayback(browserApi, tab.id, channel.platform, emit);
+          await maybePrimeTabPlayback(browserApi, tab.id, channel, emit);
         }
         diagnostic(emit, "debug", `Reusing managed watch tab ${tab.id} for ${channel.username}`, channel.platform);
         return {
@@ -113,11 +113,11 @@ export async function openPinnedMutedTabWithBrowser(
         if (Object.keys(updateProperties).length > 0) {
           await browserApi.tabs.update(tab.id, updateProperties);
         }
-        if (tab.url !== channel.url) resetPlaybackPriming(tab.id);
         if (!shouldPrimePlayback(tab, channel.url, session)) {
-          resetPlaybackPriming(tab.id);
+          // Playback is healthy, so the budget has served its purpose.
+          resetPlaybackPriming(channel.platform);
         } else if (tabOptions.keepVideosUnmuted) {
-          await maybePrimeTabPlayback(browserApi, tab.id, channel.platform, emit);
+          await maybePrimeTabPlayback(browserApi, tab.id, channel, emit);
         }
         diagnostic(emit, "debug", `Reusing your tab ${tab.id} for ${channel.username}`, channel.platform);
         return { tabId: tab.id, managedByExtension: false };
@@ -135,7 +135,6 @@ export async function openPinnedMutedTabWithBrowser(
     if (!browserApi.tabs.remove) continue;
     try {
       await browserApi.tabs.remove(tabId);
-      resetPlaybackPriming(tabId);
       diagnostic(emit, "debug", `Removed stale watch tab ${tabId}`, channel.platform);
     } catch {
       // Stale managed tab ids should not block creating the replacement.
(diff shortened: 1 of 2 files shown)


jamezrin:
Revised after real user diagnostics: the first version of this fix would **not** have helped the reported case.

The log shows the scheduler condemning the watch tab and recreating it, so the tab id changes on every cycle (`…654` -> `…669` -> `…673`):

```
06:25:12  Managed watch tab was closed; re-running scheduler
06:25:14  Switching watch target (Watch tab playback did not become active); playback unhealthy
06:25:14  Managed watch tab 1450140669 is gone; opening a new one
06:25:14  Priming playback on watch tab 1450140673
06:25:14  Opened watch tab 1450140673 for xqc
06:25:14  No video element found in the watch tab
06:25:16  Playback telemetry: 1/1 videos playing
```

The budget was a `Map<number, PlaybackPrimeState>` keyed by tab id, and it was explicitly reset on tab creation, so each churn cycle got a brand-new budget of 3 and the cap never engaged.

**What changed:** the budget is now `Map<Platform, PlaybackPrimeState>` and the state carries the `channelUrl` it was accrued for. The thing being rate-limited is "how often we yank focus to prime playback for this watch target", not "how often we prime this tab object".

- A replacement tab for the **same** channel keeps spending the same budget — the reset on tab creation and on stale-tab removal is gone, and its absence is commented so it is not re-added by accident.
- A tab for a **different** channel does not match the stored `channelUrl`, so it legitimately starts a fresh budget.
- The only reset is genuine recovery: a tick where `shouldPrimePlayback` is false.
- Back-off (`PLAYBACK_PRIME_BACKOFF_MS`, 5 min), the cap (`PLAYBACK_PRIME_MAX_ATTEMPTS`, 3) and the single `warn` diagnostic are unchanged; the warning now names the channel rather than the churning tab id.

**New regression tests**, both verified failing against the previous commit before the fix:

- *stops priming across watch tab recreation for the same channel* — `tabs.get` always rejects and `tabs.create` returns a new id each cycle, mirroring the log. Failed with `expected 6 activations to be 3`; now exactly 3 activations and 1 warn across 6 cycles.
- *primes a recreated watch tab again when the channel changed* — failed with `expected 4 to be 3`; now an exhausted budget for one channel does not suppress priming for the next.

`pnpm check` exit 0: 6/6 typechecks Done, 47 extension files / 870 tests, 9 CLI files / 125 tests, script tests 17 + 70, 0 failures, site build complete. Core stays browser-free and `coreBoundary.test.ts` is green. `controller.ts` / `handleTabRemoved` untouched (#249).

Noted and not acted on: the underlying churn (health check condemning the tab ~2s before the player reports playing) is being filed separately; this PR bounds the focus stealing only.

Issue #242, opened 2026-07-24
fix(kick): watch tab steals focus continuously (ad detection false positive)


jamezrin:
Reported by a user via email (Kick farming, RU locale): the farming tab is *so aggressive* that it becomes impossible to click the browser's extensions menu to disable LurkLoot — "you have to strain and work faster than the script, or open a new browser and disable it there".

This is **not** the reopen loop fixed in #193 / 04413e0. It is the ad-focus path repeatedly bringing the Kick watch tab to the foreground and focusing its window.

## Root cause

`packages/extension/src/core/playbackContent.ts` uses a substring ad selector for Kick:

```ts
kick: [".vjs-ad-playing", ".vjs-ad-loading", '[class*="ad-overlay"]', '[data-testid*="ad"]'],
```

`[data-testid*="ad"]` matches any testid containing the letters "ad" — including Kick's chat **`identity-badge-*`** elements (b-**ad**-ge). Verified live on `https://kick.com/asmongold`:

```
[data-testid*="ad"] -> 2 matches: identity-badge-subscriber, identity-badge-chat_identity
detectAd("kick") === true
```

Every Kick channel with chat has these, so `adActive` is permanently true. The chain then fires continuously:

1. `playbackContent.ts` reports telemetry every 5s **and** on every DOM mutation (250ms debounce). Kick chat mutates constantly, so this is roughly 4x/second.
2. `controller.ts:996` forwards `adActive` to `applyAdFocus` on every telemetry message while watching.
3. `core/tabs.ts:263-266` calls `tabs.update(tabId, { active: true })` and `windows.update({ focused: true })` — `adFocusMode` defaults to `"window"`.

Two aggravating defects in `applyAdFocusWithBrowser` (`packages/core/src/core/tabs.ts`):

- **Not idempotent** — `alreadyHeld` only gates the debug log; the activate + window-focus calls are re-issued on *every* telemetry report, including for genuine Twitch ads.
- **The hold never releases** — `releaseAdFocus` only runs when `adActive` flips false, which never happens on Kick, so the user's previous tab/window is never restored.

## Expected

Focus is only taken for a real ad, at most once per ad, and is always given back.

## Scope

- Tighten the Kick ad selectors so chat badges and similar substrings cannot match.
- Make `applyAdFocusWithBrowser` a no-op when the hold is already held and the tab is already active/focused.
- Bound the hold so a stuck detector cannot pin focus indefinitely.

PR #254, opened 2026-07-25
fix(scheduler): grant a new watch tab a playback grace period

### LINKED ISSUE #250: fix(scheduler): playback health condemns a watch tab before the player can load
Root cause behind #242 / #246 / #247. Confirmed by a real user diagnostic log (Kick, 1.8.1, RU user on a VPN with high RTT).

## Evidence

```
06:25:12  Managed watch tab was closed; re-running scheduler
06:25:14  Switching watch target (Watch tab playback did not become active); playback unhealthy
06:25:14  Managed watch tab 1450140669 is gone; opening a new one
06:25:14  Priming playback on watch tab 1450140673
06:25:14  Opened watch tab 1450140673 for xqc
06:25:14  No video element found in the watch tab
06:25:14  Playback telemetry: 0/0 videos playing
06:25:14  Playback telemetry: 0/1 videos playing
06:25:16  Playback was blocked for 1 video(s); re-muted to keep farming
06:25:16  Playback telemetry: 1/1 videos playing
```

Tab ids across the episode: `1450140654` -> `1450140669` -> `1450140673`. The extension destroys and recreates the watch tab in a loop.

Note the timing: the tab reports `0/0 videos playing` and "No video element found" immediately after opening, and reports `1/1 videos playing` **two seconds later**. Playback was fine. The health check simply looked too early.

Every Kick request in the same log reads `service worker OK (tabless-capable)` — the user's API access is healthy. This is not a blocked-platform fallback.

## Root cause

`isPlaybackHealthy` (`packages/core/src/core/scheduler.ts:1072`) requires `videoCount > 0 && playingVideoCount > 0`, with no allowance for a tab that was just created. `evaluateKeep` increments `playbackChecks` whenever that returns false and switches target once it reaches `settings.offlineRetryLimit` (default 3, `shared/settings.ts:70`).

A player that is slow to attach — slow network, high-latency VPN, a heavier page, a cold cache — can report `0` videos for the first ticks after tab creation. The scheduler condemns the tab, opens a new one, and the new tab starts from the same cold state, so the loop is self-sustaining.

## Why it matters beyond wasted cycles

Each recreated tab re-triggers `primeTabPlayback`, which foreground-activates the tab (#246), and a fresh page re-triggers the ad-focus path (#242). The churn is what turns those two into a continuous, browser-hijacking experience — the user could not reach the extensions menu to disable LurkLoot. Fixing the churn removes the driver; #245/#248 only bound the symptoms.

## Scope

- Give a newly created or newly navigated watch tab a grace period before `playbackChecks` starts accumulating — long enough for a slow player to attach. Base it on tab/session creation time, not on tick count.
- Consider making a health *recovery* reset `playbackChecks` promptly, so a tab that flickers unhealthy then recovers is not condemned by stale counters.
- Consider a floor on how often the scheduler may destroy/recreate a watch tab for the same target, so no future health signal can produce this loop again.
- Keep core browser-free; deterministic unit tests with mocked adapters and injected clocks.

Related: #242 (merged, #245), #246 (PR #248), #247 (PR #249). PR #249 also edits `scheduler.ts` — coordinate to avoid conflicts.

jamezrin:
## Summary

The scheduler condemned a watch tab whose player had not attached yet, destroyed it, and opened an equally cold replacement — a self-sustaining loop. This gives a newly opened or re-navigated watch tab a grace period before unhealthy playback checks accumulate, and makes both a tab replacement and a confirmed playback recovery clear the counter.

`Closes #250`

## Root cause

`isPlaybackHealthy` (`packages/core/src/core/scheduler.ts`) required `videoCount > 0 && playingVideoCount > 0` with no allowance for tab age. `shouldKeepWatching` incremented `playbackChecks` on every false and switched target at `settings.offlineRetryLimit` (default 3).

From the user log in #250 (Kick, 1.8.1, high-RTT VPN):

```
06:25:14  Opened watch tab 1450140673 for xqc
06:25:14  No video element found in the watch tab
06:25:14  Playback telemetry: 0/0 videos playing
06:25:16  Playback telemetry: 1/1 videos playing
```

Playback was fine two seconds later — the health check simply looked too early. Tab ids `1450140654 -> 1450140669 -> 1450140673` show the recreate loop. Health ticks also run off-cycle (tab closed, telemetry arriving), so several of them landed inside those first seconds.

## User impact

Each recreated tab re-ran `primeTabPlayback` (foreground activation, #246) and re-triggered the ad-focus path (#242), so the churn is what turned those into a continuous browser hijack — the reporter could not reach the extensions menu to disable LurkLoot. #245 and #248 bound the symptoms; this removes the driver.

## Changes

- `packages/shared/src/models.ts` — new `WatchSession.watchTabOpenedAt`.
- `packages/core/src/core/scheduler.ts`
  - `WATCH_TAB_PLAYBACK_GRACE_MS = 90 * 1000` and `isWithinPlaybackGrace`.
  - `nextPlaybackChecks` replaces the two duplicated increments (main keep path and the Idle Watchlist fallback path); it returns 0 while the session is tabless, healthy, **or** inside the grace window.
  - `watchTabOpenedAt` is stamped when the tab is fresh (new tab id, different channel, or a switch out of tabless) and carried otherwise; cleared on idle and on tabless.
  - A fresh replacement tab now starts with `playbackChecks: 0` — the old counter described the tab that was just discarded, so the replacement previously started one bad check from being killed again.
  - Extracted `isPlaybackTelemetryHealthy` so the controller can share the predicate.
- `packages/core/src/background/controller.ts` — `recordPlaybackTelemetry` resets `playbackChecks` the moment telemetry confirms playback, so recovery lands between ticks rather than waiting for the next one.

Health semantics are unchanged: the same telemetry is still required, just not before the page has had a chance to produce it.

## Why 90 seconds

`pollIntervalMinutes` defaults to 1, so 90s is one full poll interval plus slack. A freshly opened tab is therefore never condemned by its first scheduled evaluation, nor by the off-cycle ticks that caused the reported loop, even on a slow/high-latency connection. A tab that genuinely never plays is still replaced after `offlineRetryLimit` checks past the window. The window is measured from tab creation, not a tick count, because ticks are not evenly spaced.

## Recreate-rate floor

Not implemented as a separate mechanism, deliberately. The grace window plus the counter reset already give an effective floor: a managed watch tab for a given target cannot be destroyed less than `WATCH_TAB_PLAYBACK_GRACE_MS` plus `offlineRetryLimit` unhealthy checks after it was opened, and that floor rearms on every replacement. A standalone rate limiter would need its own state and would suppress legitimate replacements (channel offline, target switch) that must stay immediate. Happy to add one if a reviewer disagrees.

## Interaction with the open PRs

- **#249 (`feat/manual-close-pauses-farming`)** — also edits `scheduler.ts`, but at the top of `runSchedulerTick` (a `manualClosePause` gate) rather than in the health logic; this diff is confined to `shouldKeepWatching` / `sessionForDecision` / the tab branch. Conflicts should be trivial or none. One note for whoever merges second: #249's pause block resets `playbackChecks: 0` and does not clear `watchTabOpenedAt`; that is harmless (resuming prepares a tab, which restamps it), but adding `watchTabOpenedAt: undefined` there is tidier.
- **#248 (`fix/playback-prime-focus-steal`)** — edits `packages/core/src/core/tabs.ts`, untouched here. Complementary: #248 stops a prime from stealing focus, this stops the repeated primes from happening at all.

## Testing

`pnpm check` — pass.

```
packages/cli test:        Test Files  9 passed (9)   Tests  125 passed (125)
packages/extension test:  Test Files 47 passed (47)  Tests  869 passed (869)
```

Written TDD; all three new scheduler tests and the controller test were confirmed failing before the implementation.

- `scheduler.test.ts` — "does not condemn a freshly opened watch tab whose player has not attached yet (#250)": replays the log, four off-cycle ticks with `0/0 videos` on a tab opened seconds ago, asserts no `watch_unhealthy`, `playbackChecks` stays 0, tab id is stable, and the tab is kept once telemetry reports `1/1`.
- `scheduler.test.ts` — "still replaces a watch tab that never plays once the grace period has elapsed": a tab opened 10 minutes ago that still reports 0 playing is condemned and replaced. This is the guard that the health check was not weakened.
- `scheduler.test.ts` — "stamps the watch tab open time on a new tab and keeps it while the tab survives".
- `backgroundController.test.ts` — "clears accumulated playback checks as soon as the watch tab reports playback (#250)".
- The existing "reloads the watch tab after repeated playback failures" test still passes; its incidental `playbackChecks` assertion changed from 3 to 0 to match the replacement-tab reset described above.

`coreBoundary.test.ts` is green — no browser globals added; tests use `vi.setSystemTime` rather than sleeping.


jamezrin:
@coderabbitai review

PR #265, opened 2026-07-25
feat(scheduler): configure tabless fallback failure threshold

### LINKED ISSUE #255: feat(scheduler): configurable failure threshold before falling back to tab watching
Today the tabless -> tab fallback is effectively hair-triggered and shares a setting with an unrelated concern. Give it its own configurable failure threshold, with one deliberate exception.

## Current behaviour

`chooseTablessWatch` (`packages/core/src/core/scheduler.ts:55-66`) drops to a tab when:

```ts
if (sameChannel && previous.watchMode === "tabless" && (previous.heartbeatChecks ?? 0) >= settings.offlineRetryLimit) return false;
```

Two problems:

1. **The threshold is borrowed.** `offlineRetryLimit` (default `3`, `packages/shared/src/settings.ts:70`) exists to decide when a channel is really offline. Reusing it for "how many heartbeat failures before we give up on tabless" couples two unrelated tuning decisions — you cannot make the fallback more patient without also making offline detection more patient.
2. **It is not user-configurable as its own knob.** A transient heartbeat failure (rate limit, a blip, high-RTT VPN) costs the user a foreground tab they explicitly opted out of by enabling tabless mode. Users on constrained setups want to say "try harder before opening a tab"; users who want maximum reliability want the opposite.

Once `tablessFallback` is set, the session stays on the tab for that channel (`scheduler.ts:62`, `:751`), so a premature fallback is sticky until the target changes.

## Proposed change

- Add a dedicated setting, e.g. `tablessFallbackFailureLimit`, defaulting to something more forgiving than the current effective `3` (suggest `5`, to be confirmed against real heartbeat failure rates).
- Use it in `chooseTablessWatch` in place of `settings.offlineRetryLimit`; leave `offlineRetryLimit` to mean only "channel looks offline".
- Normalize/clamp it alongside the other numeric settings (`packages/shared/src/settings.ts:192`).
- Surface it in the popup Settings view under the existing tabless/advanced grouping, with a localized label + description in `packages/locales/messages/`.
- Existing users must keep working without a stored value — the normalizer default covers this, no migration.

## Explicit exception: Client-Integrity minting

**Do not apply the threshold to the Client-Integrity capture path.**

`ensureTwitchIntegrityWithBrowser` (`packages/core/src/core/tabs.ts:411`) opens a twitch.tv page-context tab when there is no valid token, because authenticated GQL mutations (drop claims) cannot carry a valid `client-integrity` header without one. That is not a retryable failure: without a token the request will *never* start succeeding on its own, so retrying N times before opening the tab is pure added latency on every claim, and risks missing claim windows.

This path must keep its current behaviour: no threshold, no retry budget, open the tab immediately on the first miss. It is a distinct mechanism from the watch-tab fallback (page-context tab, tracked in `managedPageContextTabs`, not a watch session), so the change should not touch it — but call it out in the PR and add a test asserting the integrity path still opens a tab on the first miss, so a future refactor that unifies "needs a tab" logic cannot quietly regress it.

## Acceptance

- New setting exists, is normalized, is persisted, and is exposed in the popup with localized copy.
- `chooseTablessWatch` no longer reads `offlineRetryLimit`.
- Tests cover: below threshold stays tabless, at threshold falls back, the setting is honoured at non-default values, and `ensureTwitchIntegrityWithBrowser` still opens a tab on the first miss with no threshold applied.

jamezrin:
## Summary

- add a dedicated `tablessFallbackFailureLimit` engine setting, defaulting to 5 and clamped from 1 to 10
- use the new threshold consistently in the heartbeat controller and scheduler while leaving `offlineRetryLimit` for channel/playback health
- expose the threshold in Advanced popup settings with localized copy across all supported locales
- preserve immediate Twitch Client-Integrity page-context acquisition and cover it with an explicit regression test

## Impact

Users can choose how patient tabless farming should be before Lurkloot opens a sticky watch tab. Existing users require no migration because missing values normalize to the new default.

## Testing

- `pnpm verify`
- 979 extension tests and 126 CLI tests passed
- workspace typechecks passed
- site, Chromium extension, and Firefox extension builds passed
- task-scoped and whole-branch code reviews reported no findings

The existing Vite large-chunk advisory remains non-fatal and unchanged.

Closes #255

### DIFF
--- a/docs/superpowers/plans/2026-07-26-tabless-fallback-failure-limit.md
+++ b/docs/superpowers/plans/2026-07-26-tabless-fallback-failure-limit.md
@@ -1,0 +1,481 @@
+# Tabless Fallback Failure Limit Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Add a persisted 1–10 setting, defaulting to 5, that independently controls how many consecutive tabless heartbeat failures occur before Lurkloot opens a watch tab.
+
+**Architecture:** The setting belongs to the shared `EngineSettings` contract so every core host receives the same normalized policy. The background heartbeat loop triggers a scheduler tick at the threshold, and the scheduler uses the same threshold to choose the sticky tab fallback; `offlineRetryLimit` remains exclusive to channel and playback health. The popup exposes the value through the existing advanced settings registry, while Client-Integrity capture remains an immediate, separately tested page-context operation.
+
+**Tech Stack:** TypeScript 7, React 19, Vitest, pnpm workspace, JSON locale catalogs.
+
+## Global Constraints
+
+- Name the setting `tablessFallbackFailureLimit`.
+- Default to `5`; normalize to an integer from `1` through `10`.
+- Do not add a settings migration; missing values must default through `mergeEngineSettings`.
+- Do not change heartbeat cadence, offline detection, playback-health thresholds, or sticky fallback semantics.
+- Do not change `ensureTwitchIntegrityWithBrowser`; a missing token must open or reuse Twitch page context immediately.
+- Keep all eleven locale catalogs in key parity.
+- Follow test-driven development and run `pnpm verify` before completion.
+
+---
+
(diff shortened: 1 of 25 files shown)


jamezrin:
@coderabbitai review

jamezrin:
@coderabbitai review

PR #282, opened 2026-07-26
fix(twitch): recover rejected integrity tokens across authenticated operations

### LINKED ISSUE #281: fix(twitch): recover rejected integrity tokens across authenticated operations
## Summary

Authenticated Twitch operations in the browser extension can be rejected with `failed integrity check` even while authentication health is `healthy`. Recovery is inconsistent:

- drops-dashboard discovery never acquires a fresh integrity token and immediately falls back to retained campaigns;
- drop claiming appears to refresh, but calls a non-forcing fast path that can reuse the same locally unexpired token Twitch rejected;
- other authenticated reads and the channel-points mutation have no integrity recovery.

Users should not have to manually open or reload `https://www.twitch.tv/drops/inventory`. LurkLoot already has a browser-backed mechanism capable of opening an inactive Twitch page context and capturing a page-minted Client-Integrity token; authenticated operations need a shared, bounded way to request a genuinely fresh token.

## Confirmed production reproduction

Observed in extension version 1.9.0 on Chrome 150 / Linux:

```text
[info] twitch authentication health changed from checking to healthy
[warn] Twitch drops dashboard request failed; reusing the last campaign list it returned: failed integrity check
```

The warning repeats approximately once per scheduler tick:

```text
18:25:55 failed integrity check
18:26:55 failed integrity check
18:27:55 failed integrity check
18:28:55 failed integrity check
18:29:41 failed integrity check
```

This confirms a persistent integrity failure rather than an invalid login or a single transient request.

## Confirmed root cause

Verified against `origin/develop` at `168db33`:

- `packages/extension/entrypoints/background.ts` injects `() => ensureTwitchIntegrity(emit)` into `TwitchAdapter`.
- `packages/core/src/core/tabs.ts` can acquire a Twitch page context and wait for a page-minted integrity token.
- `TwitchAdapter.fetchDashboard()` catches every non-authentication failure, logs the retained-campaign warning, and returns `{ response: {}, ok: false }` without invoking integrity acquisition.
- `TwitchAdapter.claimReward()` calls the callback again after an integrity rejection, but the callback fast-returns whenever the rejected token is still locally unexpired. Its comment says this is a forced refresh, but the callback has no force-refresh contract.
- The remaining authenticated operations call the ordinary GQL transport and either propagate, degrade, or retain stale data on integrity rejection.

## Audited operation scope

| Operation / area | Current behavior | Required policy |
|---|---|---|
| `ViewerDropsDashboard` | Retains cached campaign IDs; repeats warning every tick | Safe-read forced refresh and one retry, then retain |
| `Inventory` | Rejection aborts discovery/progress | Safe-read forced refresh and one retry |
| `DropCampaignDetails` | Retains details or omits unseen campaign | Safe-read forced refresh and one retry, then existing fallback |
| `DirectoryPage_Game` | Rejection can prevent candidate selection | Safe-read forced refresh and one retry |
| `DropsHighlightService_AvailableDrops` | Degrades to category-only validation | Safe-read forced refresh and one retry, then existing fallback |
| `VideoPlayerStreamInfoOverlayChannel` / `DropCurrentSessionContext` | Keeps older inventory progress | Safe-read forced refresh and one retry, then existing fallback |
| authenticated `CurrentUser` | Auth health becomes unavailable or watcher cannot resolve viewer ID | Safe-read forced refresh and one retry |
| `ChannelPointsContext` | Channel-points flow fails | Safe-read forced refresh and one retry |
| `DropsPage_ClaimDropRewards` | Retries, but can reuse the rejected token | Explicit forced refresh and one mutation retry |
| `ClaimCommunityPoints` | No integrity acquisition or retry | Ensure before mutation; explicit forced refresh and one retry |
| anonymous `StreamInfo` / `SearchCategories` | Deliberately use `credentials: "omit"` | Remain anonymous; never acquire integrity |
| Twitch heartbeat telemetry | Separate telemetry transport | No generic replay |
| Kick / non-web Twitch CLI identity | Does not use Twitch web Client-Integrity | No behavior change |

Only dashboard failure is confirmed by the supplied live diagnostics. The other entries are confirmed code-path vulnerabilities; whether Twitch currently enforces integrity for each operation may vary, so tests must exercise the rejection envelope explicitly.

## Impact

- New or not-yet-started campaigns can be omitted.
- Inventory/progress refreshes can fail or remain stale.
- LurkLoot may fail to find or fully validate a channel.
- Drop and channel-points claims can fail despite a logged-in session.
- The tabless watcher can fail to resolve its viewer ID when discovery did not supply it.
- Diagnostics may repeat indefinitely without attempting recovery.

This is functional degradation, not evidence that the user's Twitch login is invalid.

### 1. Genuine forced refresh

Extend the injected callback with an explicit request:

```ts
type TwitchIntegrityRequest = { forceRefresh?: boolean };
type EnsureTwitchIntegrity = (request?: TwitchIntegrityRequest) => Promise<boolean>;
```

In forced mode:

- do not fast-return for a locally valid token;
- remember the token Twitch rejected;
- actively boot a fresh Twitch inventory page context;
- succeed only after capturing a different, valid token;
- use a fresh inactive tab or explicitly reload an extension-owned retained context;
- never navigate, close, or repurpose a user-owned Twitch tab.

### 2. Safe authenticated reads

Add a non-recursive adapter helper that:

1. sends the read once;
2. propagates authentication failures unchanged;
3. on Client-Integrity rejection only, requests `{ forceRefresh: true }`;
4. retries the identical read once when refresh succeeds;
5. propagates the final failure to the caller's existing fallback.

Apply it to the authenticated reads in the audit table. Do not put the retry indiscriminately inside the base GQL transport, because that transport also carries mutations and anonymous requests.

### 3. Explicit mutations

- Drop claims: keep proactive integrity ensure, but use `{ forceRefresh: true }` after rejection and retry once.
- Channel points: resolve context through safe-read recovery; ensure integrity before the mutation; after integrity rejection, force refresh and retry the same claim ID once.
- Never recursively retry or make more than two mutation attempts.

### 4. Existing fallbacks

Preserve retained dashboard IDs, retained campaign details, category-only validation, and inventory-only progress when refresh or the retry ultimately fails. Emit one final fallback warning with the final reason rather than duplicate warnings from each layer.

## Test plan

Add focused Vitest coverage proving:

### Integrity acquisition

- Forced refresh does not fast-return for an apparently unexpired rejected token.
- It succeeds only after a different valid token is captured.
- It creates/reloads an extension-owned inactive page context and does not modify a user-owned tab.
- Existing no-token acquisition, timeout, retention, and capture diagnostics remain intact.

### Safe reads

- Each authenticated read listed above refreshes once and retries once after `failed integrity check`.
- Successful dashboard recovery returns fresh campaigns without the retained-campaign warning.
- Refresh returning `false` and a second rejection use the operation's existing fallback without looping.
- Authentication, network, and generic platform failures do not enter integrity recovery.
- Anonymous `StreamInfo` and `SearchCategories` retain `credentials: "omit"` and never call integrity acquisition.
- The tabless watcher can recover its authenticated `CurrentUser` fallback without replaying heartbeat telemetry.

### Mutations

- Drop claims request `{ forceRefresh: true }` after rejection instead of calling the ordinary fast path twice.
- Channel-points context uses safe-read recovery.
- Channel-points mutation ensures integrity, force-refreshes after rejection, and retries exactly once.
- A second mutation rejection propagates without a third attempt.

### Verification

```bash
pnpm --filter @lurkloot/extension test -- tests/adapters.test.ts tests/tabs.test.ts tests/tablessWatch.test.ts tests/coreBoundary.test.ts
pnpm verify
```

## Acceptance criteria

- A logged-in user does not need to manually open or reload twitch.tv to recover supported authenticated operations from an integrity rejection.
- The extension may create an inactive Twitch page-context tab when Twitch requires a real web context.
- A Twitch-rejected token is genuinely replaced, even if its local expiry is still in the future.
- Safe reads receive at most one refresh and one retry.
- Mutations opt in explicitly and receive at most one retry.
- Anonymous requests never trigger integrity acquisition.
- Existing degradation/retention behavior remains available after failed recovery.
- Kick and the non-web Twitch CLI identity are unchanged.
- No new permissions, credential storage, or localized diagnostic keys are added.

jamezrin:
## Summary

Recovers automatically when Twitch rejects an authenticated request for Client-Integrity, so a logged-in user no longer has to manually open or reload `https://www.twitch.tv/drops/inventory`.

Previously, recovery was inconsistent: drops-dashboard discovery never acquired a fresh token and fell straight back to retained campaigns (repeating `failed integrity check` once per scheduler tick), drop claiming called a non-forcing fast path that could reuse the very token Twitch had just rejected, and the remaining authenticated reads plus the channel-points mutation had no recovery at all.

Closes #281

## Changes

**Genuine forced refresh** (`packages/core/src/core/tabs.ts`) — `ensureTwitchIntegrityWithBrowser` accepts a `TwitchIntegrityRequest`. In forced mode it skips the unexpired-token fast path, remembers the rejected token, and succeeds only once a *different* valid token is captured — re-capturing the same token no longer settles the wait. A new `requireFreshPageContext` page-context option makes it boot a fresh inactive tab or explicitly re-navigate an extension-owned retained context. User-owned twitch.tv tabs are skipped entirely: never navigated, reloaded, or closed.

**Bounded safe-read recovery** (`packages/core/src/platforms/twitch/index.ts`) — a new `gqlWithIntegrityRetry` helper performs one forced refresh plus one identical replay, only for Client-Integrity rejections on authenticated requests. Authentication, network, and generic platform failures keep their existing behavior. Applied to every operation in the issue's audit table:

| Operation | Entry point |
|---|---|
| `Inventory` (both attempts), `ViewerDropsDashboard`, `DropCampaignDetails` | `discoverCampaigns()` |
| `DirectoryPage_Game` | `listCandidateChannels()` |
| `DropsHighlightService_AvailableDrops` | `checkChannel()` |
| `VideoPlayerStreamInfoOverlayChannel`, `DropCurrentSessionContext` | `readProgress()` |
| `ChannelPointsContext` | `claimChannelPoints()` |
| authenticated `CurrentUser` | `checkAuthHealth()` |

The helper sits above the transport rather than inside `createTwitchGqlTransport()`, because that transport also carries mutations (which must never be replayed implicitly) and anonymous requests (which must never open a page context).

**Explicit mutation recovery** — drop claims now request `{ forceRefresh: true }` after a rejection instead of re-calling the non-forcing fast path. Channel points gained recovery it never had: safe-read context, proactive ensure before the mutation, then one forced refresh and one retry of the same claim id. Neither path recurses or makes a third attempt.

**Tabless watcher** — `resolveUserId`'s authenticated `CurrentUser` fallback recovers the same way. The anonymous `StreamInfo` lookup and the heartbeat telemetry are deliberately excluded, so heartbeats are never replayed.

**Preserved fallbacks** — retained dashboard IDs, retained campaign details, category-only channel validation, and inventory-only progress all still apply when recovery ultimately fails.

## Testing

`pnpm verify` passes clean: six package typechecks, 132 CLI tests, 1062 extension tests (28 new), the Astro site build, and both Chromium and Firefox extension builds.

New coverage in `tests/tabs.test.ts`, `tests/adapters.test.ts`, and `tests/tablessWatch.test.ts` proves:

- forced refresh does not fast-return for an apparently unexpired rejected token, and succeeds only after a genuinely different token is captured;
- it creates a fresh inactive context or reloads an extension-owned one, and never touches a user-owned tab;
- each authenticated read refreshes once and retries once, asserting the callback receives exactly `{ forceRefresh: true }`;
- a failed refresh and a second rejection fall through to the existing fallback without looping;
- anonymous `StreamInfo` / `SearchCategories` keep `credentials: "omit"`, never acquire integrity, and are never replayed;
- drop and channel-points mutations force-refresh and retry exactly once, with a second rejection propagating without a third attempt.

Existing no-token acquisition, timeout, retention, and capture diagnostics are unchanged, and `tests/coreBoundary.test.ts` still passes — `@lurkloot/core` remains free of WXT and browser globals.

## Notes for review

Two things worth a look, both deliberate:

- **`discoverCampaigns` makes two dashboard requests, not one.** The second is the `fetchRewardCampaigns` fallback issued when both lists come back empty. Each request is independently bounded to one refresh and one retry, so a fully-rejected discovery produces 2 refresh calls rather than 1. The tests assert this structure explicitly instead of weakening the per-request bound.
- **A concurrent caller can bypass forced refresh.** `acquirePageContextTab` dedupes by origin, so if another request is already holding a context for `twitch.tv`, the forced caller inherits it and `requireFreshPageContext` has no effect. This is pre-existing behavior, self-corrects on the next tick, and widening it felt out of scope here.

No new permissions, credential storage, or localized diagnostic keys. Kick and the non-web Twitch CLI identity are unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added automatic Client-Integrity recovery for authenticated Twitch reads and certain mutations, with a single bounded retry.
  * Added forced integrity refresh to guarantee replacement tokens and refreshed page context when required (including tabless viewer resolution).
* **Bug Fixes**
  * Prevented retry loops and ensured integrity-related failures surface after the bounded recovery attempts.
  * Ensured anonymous requests are never retried and never trigger integrity handling.
* **Tests**
  * Expanded coverage for authenticated/discovery/dashboard/session/claims, anonymous behavior, forced refresh replacement, and watcher recovery.
* **Documentation**
  * Added an end-to-end implementation plan for Twitch Client-Integrity retry behavior.

### DIFF
--- a/docs/superpowers/plans/2026-07-26-twitch-dashboard-integrity-retry.md
+++ b/docs/superpowers/plans/2026-07-26-twitch-dashboard-integrity-retry.md
@@ -1,0 +1,358 @@
+# Twitch Dashboard Integrity Retry Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Recover automatically when Twitch rejects authenticated extension requests for Client-Integrity, without requiring the user to open or reload twitch.tv.
+
+**Architecture:** Keep Twitch's page-minted integrity acquisition behind the injected adapter callback so `@lurkloot/core` remains browser-free. Give that callback an explicit forced-refresh mode, make the browser host obtain a genuinely new token in that mode, and centralize one bounded retry for safe authenticated reads. Keep mutation recovery explicit at each mutation so a generic transport wrapper can never repeat a non-idempotent action accidentally.
+
+**Tech Stack:** TypeScript, WXT browser APIs, Vitest, pnpm workspace
+
+## Global Constraints
+
+- `packages/core` must not import WXT or browser globals.
+- Diagnostic messages remain English literals and are not added to locale catalogs.
+- Retry only Client-Integrity rejections; authentication, network, and other platform failures retain their existing behavior.
+- Make at most one integrity refresh and one dashboard retry per dashboard request.
+- Preserve the existing retained-dashboard fallback if refresh or retry fails.
+- Safe authenticated reads may retry once; mutations must opt in explicitly.
+- Anonymous operations must remain anonymous and must never open a page context.
+- Do not store new credentials or add browser permissions.
+
+---
+
+## Audited Twitch operation scope
+
+| Operation | Policy |
+|---|---|
+| `Inventory`, `ViewerDropsDashboard`, `DropCampaignDetails`, `DirectoryPage_Game`, `DropsHighlightService_AvailableDrops`, `VideoPlayerStreamInfoOverlayChannel`, `DropCurrentSessionContext`, `ChannelPointsContext`, authenticated `CurrentUser` | Use the shared safe-read integrity recovery |
+| `DropsPage_ClaimDropRewards`, `ClaimCommunityPoints` | Ensure integrity before mutation; on an integrity rejection, force refresh and retry exactly once |
+| Anonymous `StreamInfo` and `SearchCategories` | Keep `credentials: \"omit\"`; never refresh integrity |
+| Twitch heartbeat telemetry | Keep its existing transport; do not replay heartbeats through a generic GQL retry |
+| Kick and non-web Twitch CLI identity | No behavior change |
+
+---
+
+### Task 1: Support a forced Twitch integrity refresh
+
+**Files:**
+- Modify: `packages/core/src/core/tabs.ts`
+- Modify: `packages/extension/src/core/tabs.ts`
+- Test: `packages/extension/tests/tabs.test.ts`
+
+**Interfaces:**
+- Produces: `type TwitchIntegrityRequest = { forceRefresh?: boolean }`
+- Produces: `ensureTwitchIntegrityWithBrowser(browserApi, originUrl, timeoutMs, emit, request?): Promise<boolean>`
+- Produces: `ensureTwitchIntegrity(emit?, request?): Promise<boolean>`
+
+- [ ] **Step 1: Write failing tests for forced refresh**
+
+Add focused cases under `describe("ensureTwitchIntegrityWithBrowser")` proving that `{ forceRefresh: true }` does not fast-return when an apparently unexpired token exists, obtains a token different from the rejected token, and does not close or navigate a user-owned Twitch tab. Mock the new-token capture with:
+
+```ts
+setTwitchIntegrity({
+  integrity: "rejected-token",
+  expiresAt: Date.now() + 60_000,
+});
+setTimeout(() => setTwitchIntegrity({
+  integrity: "replacement-token",
+  expiresAt: Date.now() + 60_000,
+}, { isNew: true }), 20);
+```
+
+Assert that a fresh inactive inventory page context is created or an extension-owned context is reloaded, and that the result is `true` only after `replacement-token` is captured.
+
+- [ ] **Step 2: Run the focused tests and verify they fail**
+
+Run:
+
+```bash
+pnpm --filter @lurkloot/extension test -- tests/tabs.test.ts
+```
+
+Expected: FAIL because forced refresh is not supported and the valid-token fast path returns immediately.
+
+- [ ] **Step 3: Implement forced refresh in the browser-free tab port**
+
+Add the request type and extend `ensureTwitchIntegrityWithBrowser`. In forced mode, remember the rejected token, wait specifically for a different valid token, and actively boot a fresh Twitch inventory page context. Never repurpose or navigate a user-owned Twitch tab. Preserve the current fast path when `forceRefresh` is absent:
+
+```ts
+export interface TwitchIntegrityRequest {
+  forceRefresh?: boolean;
+}
+
+const rejectedToken = request?.forceRefresh ? twitchIntegrity?.integrity : undefined;
+if (!request?.forceRefresh && hasValidTwitchIntegrity()) return true;
+```
+
+Update the waiter predicate so forced refresh succeeds only when the captured token is valid and differs from `rejectedToken`. Reuse a retained extension-owned page-context tab only if it is explicitly navigated back to `TWITCH_PAGE_CONTEXT_URL`; otherwise create a new inactive, muted tab.
+
+- [ ] **Step 4: Bind the request through the WXT adapter**
+
+Update the extension wrapper without moving browser APIs into `@lurkloot/core`:
+
+```ts
+export function ensureTwitchIntegrity(
+  emit?: EventEmitter,
+  request?: TwitchIntegrityRequest,
+): Promise<boolean> {
+  return ensureTwitchIntegrityWithBrowser(
+    browser as BrowserTabApi,
+    TWITCH_PAGE_CONTEXT_URL,
+    undefined,
+    emit,
+    request,
+  );
+}
+```
+
+- [ ] **Step 5: Run the tab tests**
+
+Run:
+
+```bash
+pnpm --filter @lurkloot/extension test -- tests/tabs.test.ts
+```
+
+Expected: PASS, including existing no-token, reused-tab, timeout, retention, and new forced-refresh cases.
+
+### Task 2: Add bounded recovery for safe authenticated Twitch reads
+
+**Files:**
+- Modify: `packages/core/src/platforms/twitch/index.ts`
+- Modify: `packages/extension/entrypoints/background.ts`
+- Test: `packages/extension/tests/adapters.test.ts`
+- Test: `packages/extension/tests/tablessWatch.test.ts`
+
+**Interfaces:**
+- Consumes: `TwitchIntegrityRequest`
+- Changes constructor callback to: `(request?: TwitchIntegrityRequest) => Promise<boolean>`
+- Produces: `TwitchAdapter.gqlWithIntegrityRetry<T>(...): Promise<TwitchGqlResponse<T>>`
+- Produces: bounded recovery for authenticated reads without changing anonymous calls
+
+- [ ] **Step 1: Write the failing dashboard adapter tests**
+
+Add tests alongside the retained-dashboard failure cases:
+
+```ts
+it("refreshes integrity and retries a rejected dashboard once", async () => {
+  let attempts = 0;
+  const ensureIntegrity = vi.fn(async () => true);
+  const fetcher = jsonFetcher((_url, init) => {
+    const op = operation(init);
+    if (op === "Inventory") return twitchInventory([]);
+    if (op === "ViewerDropsDashboard") {
+      attempts += 1;
+      if (attempts === 1) throw new Error("failed integrity check");
+      return twitchDashboard(["campaign"]);
+    }
+    if (op === "DropCampaignDetails") return twitchCampaignDetails("campaign");
+    throw new Error(`Unexpected op ${op}`);
+  });
+
+  const campaigns = await new TwitchAdapter(fetcher, ensureIntegrity).discoverCampaigns();
+
+  expect(campaigns.map((campaign) => campaign.id)).toEqual(["campaign"]);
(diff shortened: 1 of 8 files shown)


jamezrin:
Thanks — went through all three. One applied, two declined with evidence.

### ✅ Duplicated integrity classification (nitpick 2) — fixed in fe47749

Fair catch, and the plan doc had specified exactly this. Extracted a module-level `isIntegrityRejection(error, credentials?)`; the private `shouldRecoverIntegrity` is gone and all four sites now share one definition: `gqlWithIntegrityRetry`, the drop-claim catch, the channel-points catch, and `TwitchWatcher.resolveUserId` (which previously had a hand-rolled copy it couldn't share, since it can't reach an adapter private).

Incidental improvement: the drop-claim path used a bare `/integrity/i.test(message)` with no auth-rejection guard. It now inherits that exclusion, so an invalid session no longer burns a forced refresh there.

### ❌ Forced refresh bypassed for shared entries (inline comment) — declined

The suggested fix isn't implementable as written: `findOrCreatePageContextTab` is never called on the shared-entry path. `acquirePageContextTab` reuses the *existing promise*, so there is no options object to propagate `requireFreshPageContext` into — the context has already been created by the time a second caller arrives.

More importantly, the scenario it describes doesn't occur for the case that motivates it. I added a test for a burst of three concurrent forced refreshes (`tests/tabs.test.ts`, "shares one in-flight forced refresh between concurrent callers"): all three resolve `true` with **`tabs.create` called exactly once**. The origin-level dedupe in `acquirePageContextTab` means the burst boots one fresh context, and because integrity waiters are a module-level list, every caller is woken by that one context's capture. Self-contention is already handled.

The residual case — an *unrelated* page-fetch already holding an idle twitch.tv context that a forced refresh then inherits — is real, but the fix is worse than the problem: forcing a second concurrent tab for the same origin breaks the refcount and retention invariants and would trip the managed-tab churn breaker, which exists precisely to stop the reopening loop reported in the NETV4R diagnostics. On timeout it degrades to the pre-existing retained-campaign fallback and recovers next tick.

### ❌ Forced refreshes fan out across concurrent campaign details (nitpick 1) — declined

Same root cause, same evidence: the test above shows one `tabs.create` for three concurrent forced refreshes, so `Promise.allSettled` over `DropCampaignDetails` does not multiply page-context work. The adapter-level shared-promise sketch would duplicate a dedupe that already exists one layer down, and would not cover `TwitchWatcher`, which holds a separate `ensureIntegrity` reference.

The "later waiters observe the first replacement as the rejected token" concern is worth stating precisely, because it's the one part that *can* bite: a caller that enters forced refresh strictly **after** a previous refresh has fully settled captures the replacement as its own `rejectedToken` and waits for a third token. In practice concurrent rejections resolve within the same tick and join the in-flight refresh, so this needs the previous refresh to have completed first. Fixing it properly means threading the token each request actually carried back out of the transport — a much larger change than this PR, and the failure mode is one timed-out read that the next tick recovers. Added a second test ("starts a new forced refresh once the previous one has settled") to pin the sequential behavior so any future attempt at that has a baseline.

`pnpm verify` is green: six typechecks, 132 CLI tests, 1064 extension tests, site build, both browser builds.

jamezrin:
Good catch — applied in 1638c99, though not quite as suggested.

You were right that the test was weak: `resolves.toBe(true)` would pass against a cached settled promise. But `expect(browser.tabs.create).toHaveBeenCalledTimes(2)` alone wouldn't have been a real guard here — the second call only reaches `tabs.create` because `browserMock().tabs.get` returns `undefined` by default, so the retained context is judged unusable and discarded. That assertion would have been measuring a mock artifact rather than the behavior.

Strengthened it along the axis that actually matters instead. `replacement-token` is still locally valid when the second refresh starts, so the real risk is the forced path fast-returning on it. The test now proves the second call **boots its own page context and stays pending** until a genuinely different token lands:

```ts
let settled = false;
const second = ensureTwitchIntegrityWithBrowser(...).then((ok) => { settled = true; return ok; });

await vi.waitFor(() => {
  expect(browser.tabs.create.mock.calls.length).toBeGreaterThan(contextsAfterFirstRefresh);
});
expect(settled).toBe(false);

setTwitchIntegrity({ ...replacement(), integrity: "third-token" }, { isNew: true });
await expect(second).resolves.toBe(true);
```

This keeps your "a new context was created" assertion (as a relative count, so it doesn't encode how the retained context happened to be resolved) and adds the part that pins the semantics: it had not resolved before `third-token` arrived.

Mutation-tested it rather than assuming. Reverting the forced-refresh guard to `if (hasValidTwitchIntegrity()) return true;` fails 6 tests including this one, which times out in `vi.waitFor` at 1008ms instead of passing:

```
× starts a new forced refresh once the previous one has settled 1008ms
Tests  6 failed | 1058 passed (1064)
```

Source restored, `pnpm verify` green on the rebased head (which now includes the `develop` merge with #285): six typechecks, 132 CLI tests, 1064 extension tests, site build, both browser builds. Confirmed all 12 recovery call sites survived that merge intact.

Issue #293, opened 2026-07-28
fix(controller): reset queues behind an in-flight tick instead of preempting it


jamezrin:
## Problem

"Reset Lurkloot" sits on "Resetting…" for as long as whatever tick is currently running — 24s in a measured case, 65s in an earlier report. Nothing is wrong with the reset itself; it simply cannot start.

## Cause

`resetExtension` (`packages/extension/entrypoints/background.ts`) calls `controller.prepareForHostReset`, which takes `withSettingsLock(withStateLock(...))` (`packages/core/src/background/controller.ts`). `tick()` holds the state lock across the entire network-bound `runSchedulerTick`, so the reset just waits its turn.

## Why it is not simply "make the lock finer"

A reset should not *wait* for the tick at all — it should **cancel** it. The tick currently has no cancellation path: no `AbortSignal` is threaded into `runSchedulerTick`, the platform adapters, or the integrity waits, so even a finer-grained lock would leave the abandoned tick running and racing the reset's storage wipe.

## Direction

Thread a cancellation signal through the tick so a reset (and a shutdown) aborts in-flight work rather than outliving it. `abortClaimHandoffs()` already does exactly this for the post-claim handoff loop and is a reasonable model.

The same signal would fix the orphaned integrity wait described in #291.

jamezrin:
Found while investigating #287 / #289. Not addressed in #295 — filed separately rather than bundled into it.

PR #297, opened 2026-07-28
fix(twitch): stop paying Kasada's proof-of-work once per rejected operation

### LINKED ISSUE #291: fix(twitch): integrity wait times out milliseconds before the token arrives
## Problem

From a real diagnostics export:

```
21:12:16.310 [debug] Waiting up to 12000ms for a Twitch integrity token from a created page context (tab 1055180513)
21:12:28.311 [warn]  Timed out waiting for a Twitch integrity token after 12001ms (is twitch.tv logged in?)
21:12:28.311 [warn]  Twitch drops dashboard request failed; reusing the last campaign list it returned: failed integrity check
...
21:12:38.292 [info]  Captured a fresh Twitch integrity token (expires 21:42:21.063Z, in 1783s)
```

The wait gave up, the request fell back to the retained campaign list — and the token arrived **31ms later**, valid for another 29 minutes. The tick paid the full 12s, discarded the result, and degraded to stale data for nothing.

## Suspected causes

Worth investigating together rather than just raising the timeout:

1. `INTEGRITY_REFRESH_TIMEOUT_MS = 12_000` (`packages/core/src/core/tabs.ts`) may simply be too tight for a cold page-context boot. The captured token here landed at ~22s from the context opening.
2. `waitForIntegrityCapture(timeoutMs, rejectedToken)` takes no `AbortSignal`. When `probeAuthHealth`'s own 10s deadline (`controller.ts`) aborts a probe, the integrity wait and its page-context tab keep running regardless — so a timed-out probe can still be holding a context that a later caller inherits.
3. The token that arrives after the timeout **is** stored by `setTwitchIntegrity`, so the very next request would have succeeded. The failed caller could plausibly re-check before committing to its fallback.

## Impact

Every timed-out wait costs 12s and downgrades that operation to retained/stale data, and discovery issues several such requests per tick. See #291 for the amplification.

Related: #281, fixed by 439d02a, which introduced the forced-refresh path.

### LINKED ISSUE #292: fix(twitch): per-operation forced integrity refresh multiplies tick duration
## Problem

A single automation toggle produced a **24.5s** tick, measured by the new lifecycle diagnostics (#290):

```
21:12:13.719 [debug] Tick #1 started (trigger=automation_toggle, platforms=all)
21:12:14.887 [debug] GQL ViewerDropsDashboard was rejected for integrity; refreshing the token and retrying once
21:12:16.310 [info]  Opened managed page context on www.twitch.tv: reason=background_rejected
21:12:28.311 [warn]  Timed out waiting for a Twitch integrity token after 12001ms
21:12:38.262 [debug] Tick #1 finished after 24543ms (trigger=automation_toggle, platforms=all)
```

12s of that is one integrity timeout. That is the floor, not the ceiling.

## Cause

`gqlWithIntegrityRetry` (`packages/core/src/platforms/twitch/index.ts`) calls `ensureIntegrity({ forceRefresh: true })` on **every** Client-Integrity rejection, and forced mode in `ensureTwitchIntegrityWithBrowser` skips the valid-token fast path and blocks up to `INTEGRITY_REFRESH_TIMEOUT_MS` (12s) waiting for a token *different* from the rejected one, booting a fresh page context each time.

It is wired into ~10 operations, several sequential within one discovery: `Inventory` ×2 → `ViewerDropsDashboard` ×2 → `DropCampaignDetails` → `DirectoryPage_Game` → `checkChannel`. The per-request bound is "one refresh, one replay", which is correct in isolation, but there is no **per-tick** bound. A broadly-rejecting session can therefore stack multiple 12s stalls in one tick.

439d02a's own PR notes flag the related hazard: `acquirePageContextTab` dedupes by origin, so a concurrent forced caller inherits an existing context and `requireFreshPageContext` silently has no effect.

## Direction

Bound or dedupe the forced refresh **per tick** rather than per operation — one platform, one forced refresh attempt, shared by every operation in that tick. Fixing #291 (the wait that misses its token by 31ms) would independently remove most of the cost here.

Related: #281 / 439d02a introduced this path.

jamezrin:
## Summary

Closes #291
Closes #292

Filed separately, merged here because they are the same defect seen from two ends: #291 is one integrity wait timing out, #292 is that wait happening N times in a tick. Fixing either alone leaves the other worse — see *Why they merged* below.

## What the trace actually says

#291's premise is a misreading worth correcting, because it changes the fix. The issue reports the token arriving **31ms** after the wait gave up:

```
21:12:16.310  Waiting up to 12000ms … from a created page context
21:12:28.311  Timed out after 12001ms
21:12:28.311  …reusing the last campaign list
21:12:38.292  Captured a fresh Twitch integrity token
```

Timeout → capture is **9,981ms**. The 31ms is the gap to `Tick #1 finished` at `21:12:38.262` (from #292's trace) — the tick *ending*, not the timeout.

That kills #291's proposed cause 3, "the failed caller could re-check before committing to its fallback". At the timeout the token genuinely was not there and would not be for another ten seconds, and it landed after the whole tick had ended. A re-check would never have fired. Not implemented.

## Where the 22s goes

| Span | Duration | What happens |
|---|---|---|
| `14.887 → 16.310` | **1.4s** | `tabs.create` → `waitForPageContextReady` sees `status === "complete"` |
| `16.310 → 38.292` | **22.0s** | SPA boots, authenticates, solves Kasada → first GQL carrying `Client-Integrity` |

The tab was "ready" in 1.4s and produced nothing usable for another 20.6s. Two reasons, both structural:

- `waitForPageContextReady` polls for `tab.status === "complete"`, the *document* load event. For a SPA that means the HTML shell arrived — React has not booted, authenticated, or issued a request. The 12s stopwatch starts ~20s before the event it waits for can occur.
- The cost being waited on is Kasada's proof-of-work, which `twitchIntegrity.ts` already documents as the reason a token cannot be minted from the background. Being expensive is the *point* of a PoW, and it runs in a tab deliberately created `active: false`.

So 12s was not slightly tight. It budgeted for a page load while the work was a proof-of-work challenge.

## Why they merged

Raising the timeout alone makes #292 strictly worse — those stalls stack several-deep in one tick, so 12s of guaranteed-wasted wait becomes 30s. Bounding the refresh alone leaves every cold boot still timing out at 12s. The bound is what makes the longer wait affordable: **a tick now pays it once, not once per rejected operation.**

## Changes

**Boot-phase instrumentation** (`66fdbd6`) — the wait reported one aggregate duration that could not say which phase ran long. Each boundary is now stamped as crossed and reported on both outcomes:

```
Timed out … after 30000ms from a created page context
  (tab ready at 1421ms, first GQL at 8003ms, 30012ms since the tab was created)
```

The first-GQL stamp is fed by *every* `gql.twitch.tv` request, including anonymous ones carrying no `Client-Integrity` header — those are the only evidence separating "the SPA has not booted" from "the SPA is running but is still solving the challenge". Logged on success too: a 20s success is the same latency problem as a 12s timeout.

**The refresh is bounded by the rejected token** (`81266d0`) — callers capture the token *before* issuing a request and report it on rejection. A forced refresh whose rejected token has already been replaced returns the current token instead of booting another context. Concurrent callers share the in-flight refresh rather than racing into `acquirePageContextTab`, whose origin dedupe hands the later ones an inherited context that `requireFreshPageContext` cannot make fresh — the hazard 439d02a's notes flagged.

Keying on *which token was refused* rather than on elapsed time matters. I first wrote this as a 60s cooldown and it broke an existing test — `starts a new forced refresh once the previous one has settled` — which encodes Twitch rejecting the brand-new token. A time-based bound cannot tell that from the common case and would have stalled recovery for 60s. The test was right; the heuristic was wrong. It passes unchanged.

**Timeout 12s → 30s, and the probe deadline derives from it.** `DEFAULT_AUTH_PROBE_TIMEOUT_MS` was 10s against a 12s wait, so a Twitch auth probe — which reaches `gqlWithIntegrityRetry` — could *never* observe that wait succeed. It aborted first every time, and since the wait takes no `AbortSignal`, left the wait and its tab running unowned. Now `INTEGRITY_REFRESH_TIMEOUT_MS + 5_000`, so the two cannot invert again.

## Testing

`pnpm verify` clean: 6 typechecks, 132 CLI tests, 1080 extension tests, 87 script/site tests, Astro build, Chromium and Firefox builds.

Six new tests. Each was confirmed to fail against the unfixed code rather than assumed to cover it — the phase-split four by stubbing the report to `""`, the burst test by disabling the fast path:

- phase split reported on timeout, on success, with and without a first GQL, and ignoring other tabs
- **`mints once for a burst of operations all rejected on the same token`** — the #292 case: five follow-up operations after the first refresh boot zero further contexts
- **`still mints when the replacement is the token that was rejected`** — the case the cooldown got wrong

`resetTwitchIntegrityRefreshBounds()` is added to `beforeEach` alongside the existing `setTwitchIntegrity(undefined)`; the new module state leaks between tests exactly as the token does.

## Worth flagging in review

**The 30s figure is provisional.** It covers one observed sample (22s) with margin. I did not want to tune a constant against n=1, which is why the instrumentation ships first and is deliberately the larger-value half of this PR. Replace it once real distributions come in.

**The auth probe is now up to 35s.** That follows from removing the inversion, but it means a health check can boot a page context and wait out a PoW. Arguably a probe should *observe* auth rather than do 30s of work to repair it — exempting it from forced refresh would let its deadline drop back to ~10s, at the cost of reporting a stale-token session as unhealthy. Left alone here as out of scope; say the word and I will follow up.

## Not addressed

- The missing `AbortSignal` on `waitForIntegrityCapture` (#291's cause 2) is #293's cancellation work. This PR removes the *guaranteed* orphan by fixing the deadline inversion; a probe aborting for other reasons still leaves the wait running.
- `waitForPageContextReady` still waits on `status === "complete"`. Waiting for a readiness signal that means "the SPA is issuing authenticated GQL" would cut the wait further — the new first-GQL stamp is the measurement needed to justify it.

No new permissions, credential storage, or localized diagnostic keys.

### DIFF
--- a/packages/core/src/background/controller.ts
+++ b/packages/core/src/background/controller.ts
@@ -6,7 +6,7 @@ import { isFarmingActive } from "@lurkloot/shared/settings";
 import type { CompatibilityResolution, ResolvedCompatibility } from "@lurkloot/shared/compatibility";
 import { isWatchReward, reconcileCampaignAfterClaims } from "@lurkloot/shared/rewards";
 import { isPlaybackTelemetryHealthy, MANUAL_WATCH_TTL_MS, runSchedulerTick, type StopPageContextTabs } from "../core/scheduler";
-import { currentManagedPageContextTabs, registerManagedPageContextTabs, setTwitchIntegrity, syncManagedTabBreakers } from "../core/tabs";
+import { currentManagedPageContextTabs, noteTwitchGqlRequest, registerManagedPageContextTabs, setTwitchIntegrity, syncManagedTabBreakers } from "../core/tabs";
 import { dismissCriticalFailure, recordManagedTabOpen } from "../core/criticalHealth";
 import { integrityFromHeaders } from "../core/twitchIntegrity";
 import type { IntegrityHeader, TwitchIntegrity } from "../core/twitchIntegrity";
@@ -358,7 +358,12 @@ export function createBackgroundController<S extends EngineSettings = EngineSett
   // gql.twitch.tv requests. Only genuine page-minted requests carry a
   // Client-Integrity header, so integrityFromHeaders returns undefined (and we
   // ignore) our own background fetch and anonymous queries.
-  async function captureTwitchIntegrity(headers: IntegrityHeader[] | undefined): Promise<void> {
+  // `tabId` is optional so hosts that cannot attribute a request to a tab still
+  // capture tokens; it only feeds page-context boot instrumentation.
+  async function captureTwitchIntegrity(headers: IntegrityHeader[] | undefined, tabId?: number): Promise<void> {
+    // Noted before the integrity filter: an anonymous GQL request carries no
+    // Client-Integrity header but still proves the SPA has booted.
+    noteTwitchGqlRequest(tabId);
     const integrity = integrityFromHeaders(headers);
     if (!integrity) return;
     await withStateLock(() => withEventCollector(async (emit, events) => {
--- a/packages/core/src/background/controller.ts
+++ b/packages/core/src/background/controller.ts
@@ -6,7 +6,7 @@ import { isFarmingActive } from "@lurkloot/shared/settings";
 import type { CompatibilityResolution, ResolvedCompatibility } from "@lurkloot/shared/compatibility";
 import { isWatchReward, reconcileCampaignAfterClaims } from "@lurkloot/shared/rewards";
 import { isPlaybackTelemetryHealthy, MANUAL_WATCH_TTL_MS, runSchedulerTick, type StopPageContextTabs } from "../core/scheduler";
-import { currentManagedPageContextTabs, noteTwitchGqlRequest, registerManagedPageContextTabs, setTwitchIntegrity, syncManagedTabBreakers } from "../core/tabs";
+import { currentManagedPageContextTabs, INTEGRITY_REFRESH_TIMEOUT_MS, noteTwitchGqlRequest, registerManagedPageContextTabs, setTwitchIntegrity, syncManagedTabBreakers } from "../core/tabs";
 import { dismissCriticalFailure, recordManagedTabOpen } from "../core/criticalHealth";
 import { integrityFromHeaders } from "../core/twitchIntegrity";
 import type { IntegrityHeader, TwitchIntegrity } from "../core/twitchIntegrity";
@@ -57,7 +57,13 @@ function isNothingLeftToFarm(reasonCode: WatchReasonCode | undefined): boolean {
 // still transmits.
 const RECENT_HEARTBEAT_MS = 30_000;
 const PLATFORMS: Platform[] = ["twitch", "kick"];
-const DEFAULT_AUTH_PROBE_TIMEOUT_MS = 10_000;
+// Must stay strictly greater than INTEGRITY_REFRESH_TIMEOUT_MS. A Twitch probe
+// runs through gqlWithIntegrityRetry, so a rejection makes it wait on a page
+// context minting a token; when this deadline was the shorter of the two (10s
+// against a 12s wait) the probe could never observe that wait succeed. It
+// aborted first, every time, and — because the wait takes no AbortSignal (#293)
+// — left the wait and its tab running unowned behind it.
+const DEFAULT_AUTH_PROBE_TIMEOUT_MS = INTEGRITY_REFRESH_TIMEOUT_MS + 5_000;
 class AuthProbeSetupError extends Error {
   constructor(
     readonly platform: Platform,
--- a/packages/core/src/core/tabs.ts
+++ b/packages/core/src/core/tabs.ts
@@ -487,6 +487,37 @@ const INTEGRITY_REFRESH_TIMEOUT_MS = 12_000;
 // Resolvers waiting for the next captured token (see waitForIntegrityCapture).
 let integrityWaiters: Array<() => void> = [];
 
+// Phase timings for the twitch.tv page context currently being booted to mint an
+// integrity token. A cold boot costs far more than the document load: the tab
+// reports `status === "complete"` as soon as the HTML shell lands, but the token
+// only appears once the SPA has hydrated, authenticated, and completed Kasada's
+// proof-of-work (see src/core/twitchIntegrity.ts). Those phases are billed to
+// very different causes — a slow network, a slow SPA boot, or an expensive
+// challenge in a deprioritized background tab — and the aggregate wait duration
+// cannot tell them apart, so each boundary is stamped as it is crossed.
+interface TwitchContextBootTiming {
+  tabId: number;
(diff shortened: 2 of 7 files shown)


jamezrin:
Both correct — fixed in 082c577. Thanks, the first one is a genuine hole in the design rather than an oversight in the wiring.

## P2: reported token ≠ transmitted token

Confirmed exactly as described. The window is real and not narrow: `fetchTwitchInBackgroundWith` awaits two `cookies.get` calls before it reads `twitchIntegrity`, and the webRequest listener that replaces it fires independently of any of this. The failure mode is the bad one you identified — the mismatch doesn't just lose the optimisation, it *inverts* it. Sending B while reporting A makes `hasReplacementTwitchIntegrity(A)` true, so the fast path concludes someone already handled the rejection and replays B without minting. A token Twitch just refused gets retried unchanged.

Took the second of your two options — supplying it explicitly to the transport — because it removes the race rather than narrowing it. `createTwitchGqlTransport` now reads the bundle once per attempt, sets the headers itself, and stamps every `TwitchGqlFailure` it raises with the token that went out; the retry sites read `sentIntegrityToken(error)` instead of re-reading a global. Capturing at header-assembly time inside the fetcher would have worked for the envelope path, but integrity rejections also arrive as HTTP 200 with `{errors:[…]}`, and those throws happen in the outer transport function where the headers are long out of scope.

`TwitchAdapterOptions.currentIntegrityToken` became `currentIntegrity` returning the whole bundle, since the token is bound to the device and session id it was minted with and the three have to be replayed together.

## P2: forced refresh no longer requiring a fresh context

Also confirmed, and my regression — deriving `forceRefresh` from `rejectedToken != null` when I split `mintTwitchIntegrity` out. You're right that it contradicts the interface comment I wrote one function above it. The intent is now a separate parameter.

Worth noting the blast radius was wider than the comment mismatch suggests: it fires whenever nothing has been captured yet, which is precisely the cold-start case this PR is about. Those calls would inherit a user's idle tab, which issues no request, and then wait the full 30s I had just raised the timeout to.

## Tests

Both bugs get a regression test, each confirmed to fail against the specific broken behaviour rather than assumed to cover it:

- `reports the integrity token the rejected request actually carried` — `currentIntegrity` mints a distinct token per read, and the test asserts the reported token equals the `client-integrity` header of the rejected request. Fails if the value is snapshotted anywhere other than send time.
- `still requires a fresh context when no rejected token is known` — a reusable user tab is available and `tabs.create` must still be called.
- `sends the integrity trio together…` — guards the binding that motivated widening the accessor.

`pnpm verify` clean: 1083 extension tests (was 1080), 132 CLI, 6 typechecks, both browser builds.

The `active: false` / homepage-vs-inventory question from earlier is still open and unaddressed here; the first-authenticated-GQL stamp needed to settle it isn't in this PR either.

PR #305, opened 2026-07-29
perf(scheduler): run Twitch and Kick independently

### LINKED ISSUE #294: perf(scheduler): Twitch and Kick are serialized by one global state lock
## Problem

The two platforms are not independent. A slow Twitch discovery is time during which Kick does nothing, even though nothing about the two is shared.

Removing the global `running` flag (#289) fixed the *settings* coupling, and toggles now scope their tick to the platform they changed. The *execution* coupling is still there.

## What actually serializes them

Not the alarm — three things below it:

1. **One tick loops platforms sequentially.** `packages/core/src/core/scheduler.ts`:
   ```ts
   const platforms = options.platforms ?? PLATFORMS;
   for (const platform of platforms) {
   ```
   Twitch runs to completion before Kick starts, over a single shared mutable `nextState`.

2. **One global state lock.** `stateMutation` in `packages/core/src/background/controller.ts` is a single module-level promise chain guarding load→modify→persist of one `SchedulerState` blob. Two per-platform ticks would queue behind each other on it.

3. **Module-global side registries.** `registerManagedPageContextTabs` / `syncManagedTabBreakers`, the `pageContextTabs` map keyed by *origin*, and `twitchIntegrity` in `core/tabs.ts` are process-globals mirrored from state at tick entry. Two concurrent ticks would clobber each other's mirrors.

## Why a second alarm is the last step, not the first

Adding a per-platform alarm on top of this buys nothing: it would just deliver two entry points into the same global lock. Splitting the lock and the state ownership has to come first.

`tick(platforms?)` is already platform-filterable and used that way, so the plumbing partly exists.

## Direction

1. Make `SchedulerState` writes per-platform-safe (per-platform locks over per-platform slices).
2. Make the global registries per-platform.
3. Only then, a second alarm.

Worth designing before implementing — this is the largest of the outstanding items.

jamezrin:
## Summary

- fan out every multi-platform scheduler trigger into independent Twitch and Kick operations
- protect same-platform work with platform queues while merging completed platform slices through a short storage commit queue
- scope managed page-context and breaker mirrors by platform
- replace the legacy combined scheduler alarm with separate Twitch and Kick alarms, including upgrade cleanup
- run post-claim handoffs independently and preserve CLI bootstrap isolation

## Why

A slow Twitch discovery or integrity operation previously held the global scheduler/state lock and prevented Kick from doing any work. Both platforms also shared whole-state module registries, so simply adding another alarm would still have serialized or clobbered their state.

This keeps the existing `SchedulerState` storage format while making platform work independent and preserving same-platform ordering.

Closes #294.

## Testing

- `pnpm verify`
- `pnpm --filter @lurkloot/extension exec vitest run tests/coreBoundary.test.ts`
- focused red/green coverage for platform-state merging, scoped registries, overlapping scheduler execution, alarm routing, independent handoffs, and CLI bootstrap behavior

`pnpm verify` passed with 132 CLI tests, 1,176 extension tests, workspace typechecks, site checks/build, and Chrome and Firefox production builds.

## Summary by CodeRabbit

- **New Features**
  - Platform-scoped scheduler ticking with dedicated Twitch/Kick repeating alarms and safer concurrent handoff.
  - Unified campaign refresh using `refreshCampaigns`.
  - Popup automation now supports a consistent **starting** lifecycle state.
- **Bug Fixes**
  - Prevented cross-platform persistence, diagnostics, and tab/breaker updates from impacting the wrong platform.
  - Improved claim idempotency and platform-safe state merging.
- **Documentation**
  - Added/updated plans for platform concurrency, unified refresh, automation presentation, and diagnostics.
- **Tests**
  - Expanded test coverage for alarm routing, `getAlarm` lookup behavior, diagnostics correlation, and platform-specific tab retention.

### DIFF
--- a/docs/superpowers/plans/2026-07-29-current-watch-fast-path.md
+++ b/docs/superpowers/plans/2026-07-29-current-watch-fast-path.md
@@ -1,0 +1,120 @@
+# Current Watch Fast Path Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Make a steady-state Twitch scheduler tick validate the current watch once instead of scanning every eligible campaign and candidate first.
+
+**Architecture:** Derive the highest-priority campaign and active reward locally from the refreshed campaign inventory. When they match the current session, call the existing watch-retention validation before global channel selection and reuse its result. Fall back to the existing full selection path whenever the current target is no longer preferred or validation says it must switch.
+
+**Tech Stack:** TypeScript, Vitest, pnpm workspace.
+
+## Global Constraints
+
+- Keep platform behavior behind `PlatformAdapter`.
+- Keep `@lurkloot/core` free of WXT and browser globals.
+- Diagnostics are English literals and platform-scoped.
+- Preserve current campaign priority, channel exclusion, retry, playback, and fallback behavior.
+- Use deterministic tests with mocked adapters.
+
+---
+
+### Task 1: Add a steady-state retention decision
+
+**Files:**
+- Modify: `packages/core/src/core/scheduler.ts`
+- Test: `packages/extension/tests/scheduler.test.ts`
+
+**Interfaces:**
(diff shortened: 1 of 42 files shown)


jamezrin:
@coderabbitai review

PR #309, opened 2026-07-29
fix(twitch): preserve claimable reused rewards

### LINKED ISSUE #307: [Bug]: Unclaimed rewards in inventory show as claimed in Lurkloot
### What happened?

When looking at an specific campaign in Twitch, I see that it lets me claim 2 rewards but the extension shows them as already claimed.

Image

### Steps to reproduce

Unknown as of yet.

### Where are you running Lurkloot?

Chrome or Chromium extension

### Lurkloot version

1.10.0

### Browser and operating system

Version 150.0.7871.186 (Official Build) (64-bit)

### Affected streaming platform

Twitch

### Watch mode

Tabless

### Relevant logs

```shell

```

### Additional context

_No response_

### Before submitting

- [x] I searched the existing issues for this problem.
- [x] I removed passwords, cookies, tokens, authorization headers, and other credentials from this report.

jamezrin:
## Summary

- Fixes #307: rewards that reuse a benefit id from an older campaign (e.g. the recurring Overwatch "RoT S3 Lootbox") could render as already claimed even though Twitch still showed them as claimable.
- Root cause: once a completed campaign leaves `dropCampaignsInProgress`, `mergeTwitchCampaignProgress` falls back to campaign-agnostic `gameEventDrops` ownership to detect completion (needed for legacy/no-progress cases). That fallback could override a reward that was already resolved as `claimable` with a real `dropInstanceID` from the campaign-details `self` edge, because an older campaign had already earned the same reused benefit.
- Fix: when the campaign-details parse already produced a `claimable` reward with a concrete `claimId`, that explicit per-drop state now outranks the historical ownership fallback, so the reward stays claimable (and the scheduler's claim loop can still act on it) instead of being silently marked `claimed`.

## Testing

- `pnpm --filter @lurkloot/extension test -- parsers` — 57 files / 1233 tests pass, including two new regression cases: a reused-benefit reward that stays claimable after leaving v2 progress, and confirming the legitimate owned-benefit-completion fallback (no claim instance) is unaffected.
- `pnpm typecheck`

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved the claimable status of Twitch rewards when valid claim information is available.
  * Correctly marks eligible rewards as claimed when no claim instance exists, without creating an invalid claim ID.

* **Tests**
  * Added regression coverage for Twitch reward status reconciliation in campaigns no longer in progress.

### DIFF
--- a/packages/core/src/platforms/twitch/parser.ts
+++ b/packages/core/src/platforms/twitch/parser.ts
@@ -423,11 +423,15 @@ export function mergeTwitchCampaignProgress(
       // merge above can't update it. gameEventDrops is always returned, so
       // cross-check watch ownership only when the campaign-scoped v2 evidence
       // cannot answer: after a campaign leaves progress, or under legacy v1.
+      // Preserve an explicit claimable state from campaign details: it carries
+      // the current reward's real drop-instance id and outranks historical,
+      // campaign-agnostic ownership of a reused benefit.
       // Never apply it to subscription rewards or a benefit shared by several
       // tiers of this campaign (see parseTwitchReward).
       if (
         (!progress || earnedCounts === undefined)
         && merged.status !== "claimed"
+        && !(merged.status === "claimable" && merged.claimId)
         && isWatchReward(merged)
         && ownsRewardBenefit(
           (merged.benefitIds ?? []).filter((id) => !sharedBenefitIds.has(id) && !earnedBenefitIds.has(id)),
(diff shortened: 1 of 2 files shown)


jamezrin:
Superseded by #311, which fixes the same root cause (reused-benefit ownership isn't scoped by award time) without needing this PR's special-case claimable+claimId guard.

PR #322, opened 2026-07-31
feat(popup): add campaign search and fix category/reward visibility gaps

### LINKED ISSUE #306: [Feature]: Search in the campaigns view
### What problem would this solve?

It would be ideal to have a way to search for specific campaigns, so that the user can look at the status of its rewards, drag/reorder them, etc

### What would you like Lurkloot to do?

Add a button or search input to filter the campaigns by their names, games/categories or their reward names.

### Which part of Lurkloot does this affect?

Browser extension

### Alternatives considered

_No response_

### Additional context

_No response_

### Before submitting

- [x] I searched the existing issues for a similar request.
- [x] This request does not involve storing credentials, exporting cookies, or bypassing platform detection.

jamezrin:
## Summary

- Adds a search box to the popup's campaigns (Drops) view, filtering by campaign title, game/category name, or reward name (closes #306). Drag-and-drop reordering is disabled while a search query is active.
- Fixes a bug where disabling "Farm all categories" and selecting specific categories still showed campaigns from unselected categories in the Drops list, even though they were correctly excluded from farming.
- Centralizes "is this campaign farmable" logic: extracts `campaignEligibleClass`/`campaignFarmable` into `@lurkloot/shared/campaignFilters` as the single definition consumed by both the scheduler's `isEligible` and the popup's `isCampaignVisible`, replacing an independent reimplementation that had drifted from the engine's actual rules.
- Documents the resulting farmability/visibility logic with Mermaid diagrams (`docs/campaign-farmability-visibility.md`), linked from `docs/architecture.md`.

## Details

**Search** (`packages/popup-ui/src/campaignSearch.ts`, `drops.tsx`): reuses the existing diacritic-folded, multi-token `matchesSearch` helper from the settings search. Extracted a shared `SearchBox` primitive (previously settings-only) into `primitives.tsx`.

**Visibility/farmability** (`packages/shared/src/campaignFilters.ts`, `rewards.ts`, `packages/core/src/core/scheduler.ts`):

- `campaignEligibleClass(campaign, settings)` is the shared base: status/lifecycle/exclusion/category/class-eligibility/platform-link checks, ignoring reward timing.
- `campaignFarmable` = `campaignEligibleClass` + a per-reward timing check (deadline feasibility, preconditions). Used by the scheduler's `isEligible` (which layers the `priorityMode` check on top — deliberately excluded from the shared predicate, since a campaign not yet in the priority list must stay eligible-class so the user can add it).
- `isCampaignVisible` uses `campaignEligibleClass` (not the stricter `campaignFarmable`) as its main fast path, so a campaign that's momentarily un-farmable purely for reward-timing reasons stays visible instead of disappearing — that matters most for "priority list only" mode, where `campaignPriorities` is built from the currently-visible list order, so a hidden campaign could never be dragged in to become farmable.
- The category filter has no display-flag override anywhere, so it's checked first in `isCampaignVisible`, ahead of every other bucket (excluded/finished/expired/upcoming/not-linked/subscription) — it wins even for a campaign that's simultaneously in one of those other classes with its flag on.

Also moved the reward-relevance helpers (`isRewardAvailableToEarn`, `canClaimReward`, `isRewardRelevantNow`, `isRewardDeadlineFeasible`) from private functions in `scheduler.ts` into `@lurkloot/shared/rewards`, so engine and popup share one definition.

**Not in scope**: the CLI (`packages/cli/src/runtime/status.ts`) still doesn't apply any visibility filter to its status output — it logs every discovered campaign unconditionally. Left as-is since it changes CLI output and wasn't asked for; happy to follow up if wanted.

## Test plan

- [x] `pnpm --filter @lurkloot/extension test` — 1281 tests pass, including coverage for `filterCampaigns`, `campaignEligibleClass`/`campaignFarmable`, and every visibility-precedence case (category vs. excluded/lifecycle/not-linked/subscription)
- [x] `pnpm --filter @lurkloot/cli test` — 138 tests pass
- [x] `pnpm -r typecheck` — clean across all packages
- [x] `pnpm --filter site build` — site (which embeds the real popup UI) builds
- [ ] Manual popup smoke test (search box feel, category filter, drag-and-drop while searching, priority-list-only mode) not yet done — recommend a `pnpm dev` pass before merging

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -133,6 +133,8 @@ Each scheduler tick runs enabled platforms independently:
 
 Campaign ordering is shared across platforms: explicit campaign priority, platform game priority, campaign priority field, optional lowest-availability mode, ending soonest, then campaign name. Per-platform excluded drop channels filter campaign candidates only; they do not suppress Idle Watchlist fallback channels. `farmingEligibility` also narrows eligibility, through its two farming flags (`farmUnlinkedCampaigns`, `farmSubscriptionCampaigns`); the separate `dropsListFilter` is a popup view preference that affects nothing the engine does.
 
+For exactly how a campaign's farmability (`campaignFarmable`, feeding `isEligible`) and its popup visibility (`isCampaignVisible`) are decided — and why they deliberately diverge on reward timing — see [`campaign-farmability-visibility.md`](campaign-farmability-visibility.md).
+
 ## Same-Origin Fetching
 
 In the extension, most platform calls go through the page-context fetch helpers wired by `packages/extension/src/core/tabs.ts` onto abstractions from `@lurkloot/core/tabs`. They find or open a temporary tab on the platform origin, then execute `fetch` in the page `MAIN` world. This keeps requests inside the browser's normal logged-in session and any page clearance context.
--- a/docs/campaign-farmability-visibility.md
+++ b/docs/campaign-farmability-visibility.md
@@ -1,0 +1,106 @@
+# Farmability & Visibility — `packages/shared/src/campaignFilters.ts`
+
+Two questions, one shared foundation.
+
+- `campaignFarmable` answers **"will the engine actually watch this?"**
+- `isCampaignVisible` answers **"does it show in the popup's Drops list?"**
+
+Both are built from the same `campaignEligibleClass` check, so a campaign the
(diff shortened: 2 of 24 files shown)


jamezrin:
@coderabbitai review

Issue #337, opened 2026-08-01
perf(twitch): back off unavailable higher-priority campaign searches


jamezrin:
## Problem

When the current Twitch campaign is still earning but a different eligible campaign sorts above it, the current-watch fast path is skipped. Lurkloot searches the higher-priority campaign first on every scheduler tick, even when the preceding search exhaustively found no usable channel, and only then returns to the working campaign.

In a v1.10.0 diagnostics sample, every one-minute tick repeated this sequence:

- evaluate 350 candidates for `DbD: Chorus of Sin`;
- make 18 StreamInfo batch requests and 2 AvailableDrops batch requests;
- find no winning channel;
- evaluate the next campaign and continue watching `worldoftanks`;
- repeat on the next tick.

Channel selection alone took approximately 2.7–3.3 seconds per tick. The scheduler has no negative-result reuse/backoff for the unavailable higher-priority campaign.

## Where the fast path bails

`evaluatePreferredCurrentWatch()` returns `undefined` at `packages/core/src/core/scheduler.ts:1289` because the preferred (highest-sorted, active-reward-bearing) campaign is not the campaign the current session is on. The fast path is behaving as designed — it cannot know the higher-priority campaign is unusable without searching. The missing mechanism is negative-result memory, not a wider fast path.

The current watch does still win once the exhaustive search finishes: in the sample, `AMD OLC#5 EU Day 1 Drops` reports `19 candidates batch-checked, 1 candidates evaluated`, and the evaluation loop at `packages/core/src/platforms/twitch/index.ts:1033-1057` only stops early on a winner. There is no `Switching watch target` line anywhere in the sample. The whole search is discarded every minute.

## Cost split (matters for scoping)

Of the ~2.3 s spent on the unavailable campaign each tick:

- **18 StreamInfo batch requests** — liveness for all 350 ACL channels. These results are **not cached at all**; there is no channel-liveness cache in the adapter.
- **2 AvailableDrops batch requests** — availability for the ~21–40 candidates that came back live and category-matching.

Only the second group is in scope for #338. Even with a perfectly effective availability cache, the 18 StreamInfo batches would still run every tick, so **the backoff in this issue is the load-bearing change**, not the cache fix.

[...]

PR #341, opened 2026-08-01
fix(core): make campaign inventory fingerprints order-insensitive

### LINKED ISSUE #340: fix(core): make the campaign inventory change detector order-insensitive
## Problem

The campaign-inventory change detector fires on every tick, so it reports nothing.

`packages/core/src/core/scheduler.ts:797` guards two diagnostics behind a fingerprint comparison against the previous tick's persisted campaigns:

```
Campaign inventory changed (116 discovered)
13 of 116 campaigns eligible after filtering
```

In a v1.10.0 diagnostics sample both lines appear on **all 8 ticks**, always with identical 116/13 counts, over 7 minutes in which nothing about the campaign set visibly changed. A detector that always fires is equivalent to no detector: it cannot answer "when did inventory actually change?", which is the one question it exists for, and it costs an `isEligible()` pass over every campaign each tick.

The comparison is against persisted state, not an empty list — `mergePlatformState()` carries `campaigns[platform]` through the commit (`packages/core/src/background/platformState.ts:43-46`) and `runSchedulerTick` reads it back on the next tick. So the two fingerprints genuinely differ each tick.

## Two candidate causes (not yet distinguished)

1. **Order sensitivity.** `campaignDiagnosticFingerprint()` (`scheduler.ts:1267`) joins campaigns in array order. Discovery returns `[...mergedDetails, ...inventoryOnly]` (`packages/core/src/platforms/twitch/index.ts:757`), where `mergedDetails` inherits the order of the dashboard's campaign-id list. If Twitch does not return 111 ACTIVE campaigns in a stable order, the fingerprint changes while the set does not.
2. **A genuine per-tick status flip** in some campaign or reward that the counts happen to mask.

Worth settling with a one-off instrumented comparison (log the symmetric difference of the two fingerprints, not just inequality) before fixing, because it decides whether there is a second bug behind this one.

## Desired behavior

The fingerprint compares campaign *identity and status*, not the order Twitch happened to return them in. The two diagnostics fire when the inventory materially changes and stay quiet otherwise.

Making the fingerprint order-insensitive is correct regardless of which cause above holds, so it is worth doing either way — but it should not be used to paper over cause 2 if that turns out to be real.

## Acceptance criteria

- Two consecutive discoveries returning the same campaigns and statuses in a different order produce the same fingerprint and emit no change diagnostics.
- A real change — campaign added, removed, or a campaign/reward status transition — still emits both diagnostics.
- The eligibility count pass runs only when the fingerprint actually changed.
- If instrumentation shows a genuine per-tick status flip, it is filed separately rather than hidden by the fix.
- Deterministic tests cover reordering, addition, removal, and status transition.

## Related

- #337, #338 (the other per-tick items from the same diagnostics sample). Note that #337 lists "campaign inventory/eligibility materially changes" as a backoff invalidation input — that signal is only usable once this detector is accurate, so this should land first.

jamezrin:
## Summary

- canonicalize campaign inventory diagnostics by sorting campaigns and reward status fingerprints
- keep additions, removals, campaign/reward status transitions, and existing diagnostic wording intact
- avoid the eligibility-count pass when the canonical inventory is unchanged

## Root cause

The scheduler compared Twitch campaign and reward arrays in response order. Twitch can return the same inventory in a different order, so the fingerprint changed on every tick and the detector emitted diagnostics and recomputed eligibility every time.

## Validation

- focused scheduler RED/GREEN cycle: 140/140 tests passing
- extension suite: 1315/1315 tests passing
- `pnpm --filter @lurkloot/core typecheck`
- `pnpm --filter @lurkloot/extension typecheck`

Closes #340

Depends-on context: this reliable inventory signal is the invalidation input for #337.


jamezrin:
Reviewed as part of v1.12.0 milestone triage. **Recommend merging this** — the change is minimal and correct, and the defect is confirmed against current `develop`: `campaignDiagnosticFingerprint` at `packages/core/src/core/scheduler.ts:1273` still joins in array order, and discovery returns `[...mergedDetails, ...inventoryOnly]` whose order follows Twitch's dashboard response.

One gap to note rather than to fix here. #340's acceptance criteria include:

> If instrumentation shows a genuine per-tick status flip, it is filed separately rather than hidden by the fix.

This PR does no instrumentation and files nothing, so that criterion is not met by merging alone. Order-insensitivity is correct regardless of which of #340's two candidate causes holds, so I would not block on it — but the sort silences the only signal that would reveal cause 2, and #337 plans to consume this detector as a backoff invalidation input. If a campaign or reward status genuinely flips every tick, the canonical fingerprint would still change every tick and #337's backoff would be invalidated continuously — passing tests, doing nothing in production.

I've filed #359 to cover it: log the symmetric difference of the two fingerprint sets against real data for several consecutive ticks and record the result either way. It can share a capture session with #358.

Suggest editing the description to `Closes #340` plus a note that #359 carries the instrumentation criterion, so merging this doesn't read as having discharged it. Tracked under #361.

Issue #274, opened 2026-07-26
Preserve Twitch discovery retention across scheduler ticks


jamezrin:
## Problem

The Twitch discovery hardening added for the v1.10.0 release stores the last dashboard campaign IDs and campaign details in `TwitchAdapter` instance fields. However, both the extension and CLI create fresh adapters for each controller tick.

As a result, the retention cache is empty on the next tick. If the Twitch dashboard or a campaign-details request then fails transiently, discovery can return only inventory campaigns. The scheduler treats that as a successful refresh and overwrites the previously complete campaign list, causing campaigns the user has not started to disappear.

This defeats the intended transient-failure retention behavior in PR #235.

## Relevant code

- `packages/core/src/platforms/twitch/index.ts`: `retainedDashboard`, `campaignDetailsByDropId`, and their retention helpers
- `packages/extension/entrypoints/background.ts`: `createAdapters` constructs a new `TwitchAdapter`
- `packages/cli/src/transport/http.ts` and `packages/cli/src/transport/impersonate.ts`: each `createAdapters` call also constructs a new adapter
- `packages/extension/tests/adapters.test.ts`: retention tests reuse one adapter instance and therefore do not model controller ticks

## Expected behavior

A successful Twitch discovery should remain available for the configured retention window across subsequent scheduler ticks and transient dashboard/detail failures in both extension and CLI hosts.

## Acceptance criteria

- Retention state survives adapter reconstruction between controller ticks, or adapters are made genuinely persistent without introducing stale settings/event emitters.
- A controller-level regression test performs one successful tick, recreates adapters, then simulates dashboard and/or campaign-detail failure on the next tick.
- The second tick preserves previously discovered campaigns that are absent from inventory.
- A genuine successful empty dashboard response is still authoritative and does not retain obsolete dashboard campaigns.
- Extension and CLI behavior remain consistent.

PR #342, opened 2026-08-01
perf(twitch): stop refetching every campaign's details on each tick

### LINKED ISSUE #339: perf(twitch): stop refetching every campaign's details on each tick
## Problem

Every Twitch scheduler tick refetches the full campaign-details set, unconditionally.

`discoverCampaignSnapshot()` calls `fetchCampaignDetails(discoverableCampaignIds, userLogin, signal)` at `packages/core/src/platforms/twitch/index.ts:686` for every dashboard-listed ACTIVE/UPCOMING campaign, on every tick. In a v1.10.0 diagnostics sample this was:

```
Twitch campaign details finished in 1872ms (111 operations: 6 batch requests, 0 single fallbacks)
```

— repeated identically on all 8 ticks in the sample, ~1.7–2.0 s each, while the user was farming exactly one campaign. It is the single largest component of the ~3.0 s campaign refresh, and roughly a quarter of the 7.0–8.1 s tick.

There is a details cache (`TwitchDiscoveryState.campaignDetailsByDropId`, `DISCOVERY_RETENTION_TTL_MS = 30 min`), but it is **failure retention only**: `retainedCampaignDetails()` is consulted solely in the rejected-result branch (`:727-737`). It never skips a request. #274 made that retention survive adapter reconstruction; it did not make it a read-through cache, and that was correct for its scope.

## Constraints an implementation must respect

- **The payload is user-scoped, not channel-scoped.** `DropCampaignDetails` (`:129-134`) selects `user(login: $channelLogin) { dropCampaign(id: $dropID) }`, and the caller passes the *authenticated user's* login, not a broadcaster. `rememberCampaignDetails()` therefore already keys on `dropID` alone, which is sound per identity — `setAuthenticatedUser()` (`:347-353`) clears both caches when the identity changes.
- **The payload carries progress.** `TWITCH_CAMPAIGN_FIELDS` includes `self { currentMinutesWatched isClaimed dropInstanceID }` (`:114`), so details are not static metadata. A cached entry must never be served for the campaign currently being farmed.
- **Inventory already supplies the moving parts.** `mergeTwitchCampaignProgress(parsedDetails, inventory)` (`:743`) merges the per-tick inventory response over the parsed details, and inventory is fetched every tick regardless. Campaigns the user is not watching cannot accrue minutes between ticks; their details are effectively immutable until they claim or the campaign ends.
- Dashboard-driven expiry reconciliation (`reconcileInventoryCampaignStatuses`) must keep working off the fresh dashboard response, not off cached details.

## Desired behavior

Serve campaign details from the retention cache when they are fresh enough, and spend requests only on what actually changes: campaigns newly appearing on the dashboard, the campaign being farmed, entries past a (much shorter than 30 min) reuse window, and anything a claim or settings change invalidates.

Stagger or bound the refresh so a cold start does not degrade, and so the whole set does not re-expire on the same tick.

## Acceptance criteria

- A steady-state tick with an unchanged dashboard issues substantially fewer than one details operation per known campaign.
- The campaign currently being farmed always has details fresh enough that reward progress and claimability are not delayed by the cache.
- A campaign newly listed by the dashboard is fetched on the tick it appears.
- A claim, an identity change, or a relevant settings change invalidates the affected entries.
- Failure retention behavior from #274 is preserved: a transient dashboard/detail failure still cannot blank campaigns the user has not started.
- Dashboard-driven expiry reconciliation still runs against the fresh dashboard response.
- Diagnostics report served-from-cache vs fetched counts so the existing `campaign details finished in …` line stays meaningful.
- Deterministic tests cover steady-state reuse, new campaign appearance, the farmed campaign's freshness, invalidation, and service-worker restart.

## Related

- #274 (retention across ticks — the mechanism this builds on)
- #337, #338 (the other two per-tick futile-work items from the same diagnostics sample)

jamezrin:
## Summary

- Reuse successful Twitch campaign detail reads with a bounded, identity-scoped discovery cache.
- Keep the current watched/starting campaign fresh on every discovery, including after service-worker reconstruction.
- Use deterministic per-campaign freshness spread (just over three to five minutes) to avoid synchronized large-batch expiry.
- Preserve the existing thirty-minute outage fallback, dashboard reconciliation, claim invalidation, and progress merging.
- Persist only validated fresh entries at the extension storage boundary; CLI transports remain process-local.

## Reliability hardening

- Reject malformed nested detail payloads before they can poison live, retained, or persisted cache state.
- Bound restored fresh/retained deadlines to the five-minute/30-minute horizons.
- Treat conflicting detail envelopes and malformed or absent dashboard campaign collections as ambiguous.
- Keep explicit campaign absence and valid empty dashboard collections authoritative.

## Verification

- Focused boundary suite: 442 tests passed.
- Full suite: CLI 138, extension 1,350, site 3 passed.
- pnpm typecheck: all workspace packages passed.
- git diff --check: clean.
- Final task review: clean.

Closes #339

### DIFF
--- a/packages/core/src/background/platformState.ts
+++ b/packages/core/src/background/platformState.ts
@@ -79,6 +79,9 @@ export function mergePlatformState(
       source.deadlineInfeasibleRewardIds,
       platform,
     ),
+    twitchDiscovery: platform === "twitch"
+      ? source.twitchDiscovery
+      : destination.twitchDiscovery,
     lastTickAt: newestTimestamp(destination.lastTickAt, source.lastTickAt),
   };
 }
--- a/packages/core/src/core/defaults.ts
+++ b/packages/core/src/core/defaults.ts
@@ -58,12 +58,104 @@ function normalizedIsoTimestamp(value: unknown): string | undefined {
   }
 }
 
+function isOptionalString(value: unknown): boolean {
+  return value == null || typeof value === "string";
+}
+
+function isOptionalNumber(value: unknown): boolean {
+  return value == null || (typeof value === "number" && Number.isFinite(value));
+}
+
+function isOptionalBoolean(value: unknown): boolean {
+  return value == null || typeof value === "boolean";
+}
+
+function isTwitchBenefitEdge(value: unknown): boolean {
+  if (!isRecord(value)) return false;
+  const benefit = value.benefit;
+  return benefit == null || (
+    isRecord(benefit)
+    && isOptionalString(benefit.id)
+    && isOptionalString(benefit.name)
+    && isOptionalString(benefit.imageAssetURL)
+    && isOptionalString(benefit.distributionType)
+  );
+}
+
+function isTwitchCampaignReward(value: unknown): boolean {
+  if (!isRecord(value) || typeof value.id !== "string" || value.id.length === 0) return false;
+  const benefitEdges = value.benefitEdges;
+  const preconditionDrops = value.preconditionDrops;
+  const self = value.self;
+  return isOptionalString(value.name)
+    && isOptionalString(value.startAt)
+    && isOptionalString(value.endAt)
+    && isOptionalNumber(value.requiredMinutesWatched)
+    && isOptionalNumber(value.requiredSubs)
+    && (benefitEdges == null || (Array.isArray(benefitEdges) && benefitEdges.every(isTwitchBenefitEdge)))
+    && (preconditionDrops == null || (
+      Array.isArray(preconditionDrops)
+      && preconditionDrops.every((drop) => isRecord(drop) && typeof drop.id === "string")
+    ))
+    && (self == null || (
+      isRecord(self)
+      && isOptionalNumber(self.currentMinutesWatched)
+      && isOptionalBoolean(self.isClaimed)
+      && isOptionalString(self.dropInstanceID)
+    ));
+}
+
+function isTwitchAllowedChannel(value: unknown): boolean {
+  return typeof value === "string" || (
+    isRecord(value)
+    && isOptionalString(value.name)
+    && isOptionalString(value.login)
+  );
+}
+
+function isTwitchCampaignGame(value: unknown): boolean {
+  return value == null || (
+    isRecord(value)
+    && isOptionalString(value.id)
+    && isOptionalString(value.name)
+    && isOptionalString(value.displayName)
+    && isOptionalString(value.slug)
+    && isOptionalString(value.boxArtURL)
+  );
+}
+
 export function isTwitchCampaignDetailPayload(value: unknown, dropID: string): value is Record<string, unknown> {
-  return isRecord(value)
-    && value.id === dropID
-    && typeof value.name === "string"
-    && value.name.length > 0
-    && Array.isArray(value.timeBasedDrops);
+  if (
+    !isRecord(value)
+    || value.id !== dropID
+    || typeof value.name !== "string"
+    || value.name.length === 0
+    || !Array.isArray(value.timeBasedDrops)
+    || !value.timeBasedDrops.every(isTwitchCampaignReward)
+  ) return false;
+  const game = value.game;
+  const self = value.self;
+  const allow = value.allow;
+  const allowedChannels = value.allowedChannels;
+  return isTwitchCampaignGame(game)
+    && isOptionalString(value.imageURL)
+    && isOptionalString(value.startAt)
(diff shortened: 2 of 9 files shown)


jamezrin:
Reviewed as part of v1.12.0 milestone triage. **Recommend restarting this one.** The underlying issue (#339) is real and confirmed against current `develop` — `fetchCampaignDetails()` is still called unconditionally for every discoverable campaign at `packages/core/src/platforms/twitch/index.ts:686`, and `retainedCampaignDetails()` is still consulted only in the rejected-result branch. But this implementation adds a persistence layer #339 never asked for, and that layer is where both problems are.

## 1. Per-tick write amplification against a 10 MB quota

`createTwitchDiscoveryStateStorage.saveState` injects the snapshot into **every** state save:

```ts
async saveState(state: SchedulerState): Promise<void> {
  await ensureHydrated();
  const snapshot = discoveryState.snapshot();
  const nextState = { ...state, twitchDiscovery: snapshot };
  ...
  await saveState(nextState);
}
```

State saves happen once per scheduler tick, i.e. once a minute by default. The snapshot carries up to `TWITCH_DISCOVERY_SNAPSHOT_MAX_ENTRIES = 128` **raw** Twitch campaign payloads, each with its full `allowedChannels` ACL — the diagnostics sample behind #339 had a single campaign with a 350-login ACL. `saveState` writes to `browser.storage.local` (`packages/extension/src/core/storage.ts:88`), and `packages/extension/wxt.config.ts:35` declares no `unlimitedStorage` permission, so that is a 10 MB quota.

So a PR whose purpose is to remove ~1.9 s of per-tick network work adds a potentially multi-megabyte serialize-and-write on the same cadence, plus quota-exhaustion risk that would surface as state persistence failing — a much worse failure mode than a cache miss. This needs to be measured before it is defensible, and I don't think it survives measurement.

**#339 does not require persistence.** Its acceptance criterion mentioning service-worker restart is satisfied by a cold in-memory cache that refetches on the next tick — that is the pre-existing behaviour, and it is fine. `DISCOVERY_RETENTION_TTL_MS` retention from #274 already survives adapter reconstruction within a worker generation, which is the case that actually recurs.

## 2. Layering

Even setting the quota aside, the placement is wrong for this repo:

- `packages/core/src/core/defaults.ts` — the scheduler-state normalizer — grows ~150 lines of Twitch GQL structural validation (`isTwitchBenefitEdge`, `isTwitchCampaignReward`, `isTwitchAllowedChannel`, `isTwitchCampaignGame`, `isTwitchCampaignDetailPayload`). That is Twitch payload knowledge in `core/`, duplicating what `platforms/twitch/parser.ts` already encodes, with the two free to drift.
- `@lurkloot/shared/models.ts` gains `TwitchDiscoverySnapshotEntry { campaign: unknown }` — a raw platform payload smuggled through the shared contracts package as `unknown`.

CLAUDE.md is explicit: *"Keep platform behavior behind `PlatformAdapter`; do not mix Twitch/Kick parsing logic into scheduler or UI code"* and *"Put cross-package types, models, and settings in `@lurkloot/shared`"* — contracts, not payload snapshots. A `unknown`-typed Twitch GQL blob validated by a hand-rolled 150-line predicate in `core/defaults.ts` is the opposite of both.

The `TwitchDiscoveryState` changes inside `platforms/twitch/index.ts` are in the right place, and the freshness-spread idea (deterministic per-campaign jitter so a cold batch doesn't re-expire on one tick) is good and worth keeping.

## Suggested shape for the restart

Read-through cache entirely inside `TwitchDiscoveryState`, in memory:

- `retainedCampaignDetails()` becomes readable outside the rejected branch, gated on a fresh-until deadline much shorter than the existing 30-minute retention.
- Keep the deterministic freshness spread.
- Always refetch the campaign currently being farmed (`self { currentMinutesWatched isClaimed dropInstanceID }` moves), plus anything new on the dashboard this tick.
- Preserve #274's failure retention as-is.
- No `SchedulerState` field, no `models.ts` change, no `defaults.ts` validator, no storage migration.

That should be a fraction of this diff and lands the entire measured win.

## Also

The savings figure in #339 (~1.7–2.0 s/tick) comes from a v1.10.0 diagnostics sample that predates `twitch-inventory-v2` — see #358. The *defect* is confirmed against current `develop` independent of that sample, so this work is not blocked; only the quoted number should be restated from a fresh capture.

Leaving this open rather than closing it — the call on restarting versus reworking in place is yours. Tracked under #361.

Issue #301, opened 2026-07-28
[Feature]: Richer activity log


jamezrin:
### What problem would this solve?

The current activity log is very raw, akin to the diagnostics view. The only difference is the information shown and the fact that one is localized and the other is not

### What would you like Lurkloot to do?

I want the activity log to feature rich content such as the obtained drop's image, clickable campaign/reward links, etc.

### Which part of Lurkloot does this affect?

Browser extension

### Before submitting

- [x] I searched the existing issues for a similar request.
- [x] This request does not involve storing credentials, exporting cookies, or bypassing platform detection.

Issue #300, opened 2026-07-28
[Feature]:  Search in diagnostics view


jamezrin:
### What problem would this solve?

I would like to be able to easily and quickly find diagnostic messages

### What would you like Lurkloot to do?

I would like to have a way to look for diagnostic messages containing specific things, for that I think we should add a search box that iterates through the diagnostic messages and filters for those that contain the string the user provided.

### Which part of Lurkloot does this affect?

Browser extension

### Before submitting

- [x] I searched the existing issues for a similar request.
- [x] This request does not involve storing credentials, exporting cookies, or bypassing platform detection.

PR #343, opened 2026-08-01
perf(twitch): share channel availability cache across ticks

### LINKED ISSUE #338: perf(twitch): make channel availability caching effective across scheduler ticks
## Problem

Twitch channel campaign-availability results currently use a fixed `CHANNEL_CAMPAIGN_CACHE_TTL_MS = 60_000`. The default/observed scheduler cadence is also around one minute.

Because cache lifetime and poll cadence share the same boundary—and alarms, auth probing, inventory refresh, and scheduling jitter consume time before selection—the next selection commonly reaches the cache at or after expiry.

This is stronger than a likelihood. At the default `pollIntervalMinutes: 1` the cache is **structurally incapable of hitting**:

- entries are written with `expiresAt: Date.now() + CHANNEL_CAMPAIGN_CACHE_TTL_MS` (`packages/core/src/platforms/twitch/index.ts:25`, `:1418-1421`);
- both read sites require `cached.expiresAt > Date.now()` (`:1078` batch, `:1392` single);
- the next tick reaches selection roughly 60 s later *plus* the time discovery consumes first, so the comparison always fails, and the entry is then eagerly deleted (`:1083`, `:1396`).

A v1.10.0 diagnostics sample confirms it: `0 availability cache hits` on **every** selection line across an 8-tick sample, while checking the same campaigns and the same channels each time.

This defeats the intended cross-tick cache and contributes repeated AvailableDrops traffic. Manifest V3 service-worker restarts also recreate the in-memory adapter/cache, further reducing reuse.

## Desired behavior

Give availability results a useful reuse window that is deliberately longer than the effective selection cadence, or model the cache with an explicit next-validation time rather than tying it to the nominal alarm period.

Cache policy should account for result confidence:

- successful positive and negative AvailableDrops responses may have different freshness needs;
- ambiguous results and transport failures should not become authoritative cached misses;
- bounded persistence may be needed if cross-tick reuse should survive service-worker suspension.

This issue should coordinate with #337 but remain focused on channel-level availability data and request deduplication.

Scope note: fixing this cache does **not** on its own fix #337. In the sample the unavailable campaign spent 18 StreamInfo batch requests on channel liveness (which has no cache at all) against only 2 AvailableDrops batch requests. This issue removes the smaller half; #337 removes the search.

## Acceptance criteria

- Two scheduler selections roughly 60 seconds apart can reuse unchanged, authoritative channel availability data.
- Cache expiry has intentional headroom over the configured/effective poll cadence and does not expire at the same boundary.
- Cache keys prevent reuse across the wrong Twitch identity, channel, or campaign context.
- Failed or ambiguous AvailableDrops responses do not poison negative results.
- Cache size and lifetime are bounded, including any persisted representation.
- Setting changes or authentication identity changes invalidate relevant entries.
- Diagnostics report cache hits, misses, expirations, and invalidations.
- Tests cover boundary timing, positive/negative entries, identity changes, failure handling, and service-worker reconstruction if persistence is implemented.

jamezrin:
## Stack

This PR is intentionally stacked on #342 (the finalized #339 campaign-detail cache). Its diff against the #339 branch contains only #338.

## Summary

- Move AvailableDrops caching into TwitchDiscoveryState so adapter instances share it across scheduler ticks.
- Cache complete authoritative positive and negative campaign lists for two minutes with one-minute scheduler headroom.
- Keep the cache bounded to 128 channels, identity-scoped, lazily expired, and FIFO-pruned.
- Reuse results across batch and single checks without increasing request concurrency.
- Preserve undefined for malformed, wrong-channel, incomplete, transport, or ambiguous results; auth failures still propagate.
- Guard in-flight writes with identity generations so user-switch/reset races cannot cross-contaminate results.
- Preserve anonymous StreamInfo page fallback while authenticated availability errors remain observable.

## Verification

- Adapter suite: 190 passed.
- Extension boundary suite: 459 passed.
- CLI isolation suite: 27 passed.
- Full suite: CLI 138, extension 1,367, site 3 passed.
- pnpm typecheck: passed.
- Final task review: clean.

Closes #338

### DIFF
--- a/packages/core/src/platforms/twitch/index.ts
+++ b/packages/core/src/platforms/twitch/index.ts
@@ -346,6 +346,11 @@ type CampaignAvailabilityCacheLookup =
   | { status: "hit"; available: boolean }
   | { status: "miss" | "expired" };
 
+interface AvailabilityRequestIdentity {
+  userId: string;
+  generation: number;
+}
+
 interface CachedCampaignDetails {
   campaign: unknown;
   freshUntil: number;
@@ -384,6 +389,7 @@ export class TwitchDiscoveryState {
   private readonly availableCampaignsByChannel = new Map<string, CachedAvailableCampaigns>();
   private readonly campaignDetailsByDropId = new Map<string, CachedCampaignDetails>();
   private authenticatedUserId?: string;
+  private identityGeneration = 0;
   private retainedDashboard?: CachedDashboardCampaigns;
 
   setAuthenticatedUser(userId: string): number {
@@ -393,15 +399,25 @@ export class TwitchDiscoveryState {
       this.authenticatedUserId = userId;
       return invalidatedAvailabilityEntries;
     }
+    if (this.authenticatedUserId !== userId) this.identityGeneration += 1;
     this.authenticatedUserId = userId;
     return 0;
   }
 
   clear(): void {
+    this.identityGeneration += 1;
     this.authenticatedUserId = undefined;
     this.retainedDashboard = undefined;
     this.availableCampaignsByChannel.clear();
     this.campaignDetailsByDropId.clear();
+  }
+
+  availabilityRequestIdentity(): AvailabilityRequestIdentity | undefined {
+    if (!this.authenticatedUserId) return undefined;
+    return {
+      userId: this.authenticatedUserId,
+      generation: this.identityGeneration,
+    };
   }
 
   campaignAvailability(channelId: string, campaignId: string): CampaignAvailabilityCacheLookup {
@@ -418,15 +434,23 @@ export class TwitchDiscoveryState {
     return { status: "hit", available: cached.campaignIds.has(campaignId) };
   }
 
-  rememberAvailableCampaigns(channelId: string, campaignIds: Iterable<string>): void {
-    if (!this.authenticatedUserId) return;
+  rememberAvailableCampaigns(
+    channelId: string,
+    campaignIds: Iterable<string>,
+    requestIdentity: AvailabilityRequestIdentity | undefined,
+  ): void {
+    if (
+      !requestIdentity
+      || requestIdentity.userId !== this.authenticatedUserId
+      || requestIdentity.generation !== this.identityGeneration
+    ) return;
     const now = Date.now();
     for (const [cachedChannelId, cached] of this.availableCampaignsByChannel) {
       if (cached.expiresAt <= now) this.availableCampaignsByChannel.delete(cachedChannelId);
     }
     this.availableCampaignsByChannel.delete(channelId);
     this.availableCampaignsByChannel.set(channelId, {
-      userId: this.authenticatedUserId,
+      userId: requestIdentity.userId,
       campaignIds: new Set(campaignIds),
       expiresAt: now + CHANNEL_CAMPAIGN_CACHE_TTL_MS,
     });
@@ -1392,6 +1416,7 @@ export class TwitchAdapter implements PlatformAdapter {
         ),
       );
     };
+    const requestIdentity = this.discoveryState.availabilityRequestIdentity();
     const integrity = this.options.currentIntegrity?.();
     let raw: unknown;
     try {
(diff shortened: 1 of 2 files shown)


jamezrin:
Reviewed as part of v1.12.0 milestone triage. **This one is worth salvaging** — the core idea is right and the underlying defect is confirmed against current `develop`. Two things to fix, one of which is a regression.

The defect is real: `CHANNEL_CAMPAIGN_CACHE_TTL_MS = 60_000` (`packages/core/src/platforms/twitch/index.ts:25`), both write sites use `Date.now() + TTL` (`:1228`, `:1420`), both read sites require `expiresAt > Date.now()` (batch and single), and the default poll cadence is one minute — so at the default the cache structurally cannot hit. Moving it into `TwitchDiscoveryState` so it survives adapter reconstruction, widening the TTL to two minutes for headroom, the `CHANNEL_CAMPAIGN_CACHE_MAX_ENTRIES = 128` FIFO bound, and the identity-generation guard against user-switch races are all the right calls.

## 1. Regression — per-lookup diagnostics will destroy the diagnostic history

```ts
private cachedCampaignAvailability(channelId, campaignId, metrics?) {
  const lookup = this.discoveryState.campaignAvailability(channelId, campaignId);
  diagnostic(this.emit, "debug", `Twitch availability cache ${lookup.status} for channel ${channelId}`, "twitch");
  ...
}
```

This emits one diagnostic record **per channel lookup**. The sample behind these issues evaluates ~350 candidates per tick, so that is ~350 records/minute.

The diagnostic store is a bounded ring: `MAX_RECORDS_PER_CATEGORY = 2_000` with `DIAGNOSTIC_RETENTION_MS = 7 days` (`packages/extension/src/core/activityStorage.ts:17-19`). At 350/minute the cap is reached in under six minutes, so the entire diagnostic history rolls over roughly every six minutes instead of holding seven days. That doesn't just add noise — it evicts everything else and directly regresses the diagnostic search (#300 / #334) and the activity timeline (#301) that shipped in v1.11.0. It would also make the very diagnostics-export workflow that produced #337–#340 unusable for diagnosing anything else.

**Fix:** drop the per-lookup emit entirely and keep only the aggregate counters on the existing per-selection summary line — which this PR already adds (`cacheMisses`, `cacheExpirations`). Those satisfy #338's "Diagnostics report cache hits, misses, expirations, and invalidations" criterion on their own, at one record per selection instead of 350.

While there: the summary line is now extremely long, and the pluralisation helpers (`cacheHitLabel` / `cacheMissLabel` / `cacheExpirationLabel`) add three locals to grammar-adjust a debug string. Diagnostics are English literals by convention, but `hits=N misses=N expirations=N` would read better and shrink each stored record.

## 2. Base branch

This is based on `perf/twitch-campaign-details-cache` (#342), not `develop`. I've recommended restarting #342 — see the review there — which would evaporate this base.

The good news is that nothing here actually depends on #342. The two only collide because both touch `TwitchDiscoveryState`. Re-cutting this directly against `develop` should be close to mechanical: the availability cache fields, `availabilityRequestIdentity()`, `campaignAvailability()`, `rememberAvailableCampaigns()`, and the `setAuthenticatedUser()` return-count change are all self-contained. Doing that also makes this independently reviewable and mergeable, which it currently isn't.

## Minor

`setAuthenticatedUser()` changing from `void` to `number` (invalidated-entry count) so the caller can emit a diagnostic is a slightly awkward contract — the return value means "entries invalidated" only on the identity-change path and is `0` both when nothing was invalidated and when the identity was unchanged. A separate `drainInvalidationCount()`, or emitting from inside the state object, would keep `setAuthenticatedUser` honest. Not blocking.

## Also

#338's cost framing (2 AvailableDrops batch requests against 18 uncached StreamInfo batches) comes from a v1.10.0 diagnostics sample that predates `twitch-inventory-v2` — see #358. The cache defect is confirmed independent of that sample, so this isn't blocked; only the quoted savings should be restated from a fresh capture. Worth remembering that this fixes the smaller half of #337's cost.

Leaving this open — the restart-versus-rebase call is yours. Tracked under #361.

jamezrin:
Closing to redo from scratch — restarting under a fresh branch per the v1.12.0 triage notes (#358/#337 comments have the current diagnostics baseline).

Issue #355, opened 2026-08-04
refactor(cli): dedupe Twitch cycletls heartbeat plumbing between impersonate.ts and http.ts


jamezrin:
## Problem

`packages/cli/src/transport/cycle.ts` is Kick-only by design — that's intentional, not a violation on its own (Twitch has no Cloudflare WAF in front of it, so it doesn't need the cycletls fingerprinting cycle.ts exists for).

However, the Twitch-side cycletls plumbing that *does* exist for the `impersonate` transport's heartbeat calls (`heartbeatFetchText` / `heartbeatPost` in `packages/cli/src/transport/impersonate.ts`) is inlined directly there, near-duplicated against the plain-fetch equivalents in `packages/cli/src/transport/http.ts`. Both transports build near-identical `TwitchAdapter` construction blocks (`createAdapter`, `createAdapters`) with only the heartbeat fetch implementation differing.

## Suggested fix

Factor the shared `TwitchAdapter`/`KickAdapter` construction logic (identity resolution, compatibility resolution, `createAdapter`/`createAdapters`) out of `http.ts` and `impersonate.ts` into a shared helper in `packages/cli/src/transport/common.ts`, parameterized by the platform-specific fetcher/heartbeat implementations each transport supplies. This wasn't fixed as part of #353 / #354 — flagged as a separate, lower-severity cleanup.

jamezrin:
Re-verified against current `develop` (cb7d0a4) during v1.12.0 triage — **premise still holds exactly, keeping on the milestone.**

`packages/cli/src/transport/http.ts` and `packages/cli/src/transport/impersonate.ts` both still carry their own near-identical block: `resolveCompatibility(settings.compatibility, { host: "cli", twitchIdentity })` (`http.ts:25`, `impersonate.ts:35`) followed by a `new TwitchAdapter(...)` / `new KickAdapter(...)` pair (`http.ts:27`/`:53`, `impersonate.ts:37`/`:71`) differing only in the `heartbeatFetchText` / `heartbeatPost` implementations (`http.ts:36`/`:43`, `impersonate.ts:46`/`:57`). `packages/cli/src/transport/common.ts` still holds none of it.

#351 rewrote `needsKickBearer` inside `cycle.ts` but did not touch this split, so nothing here is stale.

Two scheduling notes:

[...]

Issue #353, opened 2026-08-04
fix(cli): Kick cycletls transport reimplements core's session-bearer host check with a weaker, diverged copy


jamezrin:
## Problem

`packages/cli/src/transport/cycle.ts` (`kickHeaders`) had its own local `KICK_AUTH_HOSTS` list and attached the Kick session Bearer via `url.includes(host)`, instead of reusing `needsKickSessionBearer` from `packages/core/src/core/tabs.ts`. This diverged from the canonical predicate in two ways:

1. **Substring matching instead of parsed-host matching.** `packages/core/src/core/tabs.ts` has an explicit comment warning against this exact pattern: `.includes()` would also attach the session token to hosts that merely mention a Kick host (e.g. `https://evil.example/?r=web.kick.com`).
2. **Missing the `kick.com` + `/api/v1/user` case.** Core's predicate treats `kick.com/api/v1/user` as needing the Bearer too (Kick serves it anonymously as `200 {}` instead of a 401/403), but `cycle.ts`'s list only had `web.kick.com` and `websockets.kick.com`. This meant `KickAdapter.checkAuthHealth()`'s identity probe never got its Bearer token when running over the CLI's `impersonate` (cycletls) transport, so a valid session would look permanently signed out.

## Fix

Export `needsKickSessionBearer` from `packages/core/src/core/tabs.ts` and use it directly in `cycle.ts`'s `kickHeaders`, removing the local `KICK_AUTH_HOSTS` reimplementation.

Already fixed on branch `fix/kick-transport-adapter-leak` (in `.worktrees/kick-transport-adapter-leak`), pending a PR into `develop`.

jamezrin:
Update: this was independently fixed by #351, which landed on \`develop\` (commit cb7d0a4) while this was in progress. #351's version is more correct than what I'd drafted — it exact-matches host+pathname (not substring), and additionally handles the \`wss:\` protocol for the viewer WebSocket case, which my original fix (delegating straight to core's \`needsKickSessionBearer\`) would have broken (that function only accepts \`https:\`, and \`cycle.ts\`'s \`kickHeaders\` is also used to build the WebSocket handshake headers via \`createCycleKickWebSocketFactory\`).

Rebased my branch on latest \`develop\` and dropped the auth-host predicate change entirely — closing as fixed by #351.

PR #362, opened 2026-08-04
fix(cli): classify Kick cycletls 403s the same way checkAuthHealth expects

### LINKED ISSUE #354: fix(cli): Kick cycletls transport misclassifies non-WAF 403s as KickWafBlockedError over the impersonate transport
## Problem

`packages/cli/src/transport/cycle.ts` (`createCycleKickFetcher`) treated *every* HTTP 403 from Kick as `KickWafBlockedError`, and every other `>=400` status as a plain `Error`. This diverged from the canonical classifier core already uses for the same decision.

`packages/core/src/core/tabs.ts`'s `fetchKickInBackgroundWith` (used by the extension's page-context fetcher) instead calls `safeKickFailure(status, text)`, which inspects the response body to distinguish:
- `security_policy_blocked` (Cloudflare WAF challenge) → `KickWafBlockedError`
- `authentication_rejected` (bad/expired session) → plain `SafeFetchError`

`KickAdapter.checkAuthHealth()` (`packages/core/src/platforms/kick/index.ts`) depends on this distinction to report `invalid_credentials` vs `blocked` vs `unavailable`. Because `cycle.ts`'s fetcher only ever threw a string-constructed `KickWafBlockedError` (which always resolves to `failure.kind: "network_error"`), `checkAuthHealth()` could never surface `invalid_credentials` or `blocked` over the CLI's `impersonate` transport — every failure showed up as the generic "platform unavailable" / "network unavailable" state, regardless of whether the actual problem was a WAF block or a genuinely bad/expired Kick session token.

## Fix

Export `safeKickFailure` from `packages/core/src/core/tabs.ts` and mirror `fetchKickInBackgroundWith`'s classification in `cycle.ts`. Also added a guard for the case where Cloudflare's block against the impersonated cycletls session comes back as a non-JSON HTML challenge page rather than Kick's JSON error envelope (which `safeKickFailure`'s body-regex match alone wouldn't catch) — a 401/403 with an unparseable body is treated as `KickWafBlockedError` directly.

Already fixed on branch `fix/kick-transport-adapter-leak` (in `.worktrees/kick-transport-adapter-leak`), pending a PR into `develop`.

jamezrin:
## Summary

Closes #354.

`createCycleKickFetcher` (the Kick `PageFetcher` used by the CLI's `impersonate` transport) treated **every** HTTP 403 from Kick as `KickWafBlockedError`, and every other `>=400` status as a plain `Error`. This diverged from the classification core already uses for the same decision.

`fetchKickInBackgroundWith` (`packages/core/src/core/tabs.ts`, used by the extension's page-context fetcher) instead calls `safeKickFailure(status, text)`, which inspects the response body to distinguish `security_policy_blocked` (Cloudflare WAF challenge) from `authentication_rejected` (bad/expired session). `KickAdapter.checkAuthHealth()` depends on this distinction to report `invalid_credentials` vs `blocked` vs `unavailable`. Because the CLI's cycletls fetcher only ever threw a string-constructed `KickWafBlockedError` (which always resolves to `failure.kind: "network_error"`), `checkAuthHealth()` could never surface `invalid_credentials` or `blocked` over the `impersonate` transport — every failure showed up as the generic "unavailable" state, regardless of whether the real problem was a WAF block or a genuinely bad/expired Kick session token.

## Changes

- Export `safeKickFailure` from `packages/core/src/core/tabs.ts` (mirrors the existing `KickWafBlockedError` export, same rationale: reused by the CLI's headless transport).
- `cycle.ts`'s `createCycleKickFetcher` now classifies `>=400` responses the same way `fetchKickInBackgroundWith` does.
- Added a guard for the case where Cloudflare's block against the impersonated cycletls session comes back as a non-JSON HTML challenge page rather than Kick's JSON error envelope (which `safeKickFailure`'s body-regex match alone wouldn't catch): a 401/403 with an unparseable body is treated as `KickWafBlockedError` directly.

## Not in scope

The other architecture-review finding from the same pass (`KICK_AUTH_HOSTS`/session-bearer predicate duplication, #353) was independently fixed by #351 before this branch was rebased on top of it, so it's not part of this diff. A milder, separate finding (Twitch cycletls plumbing duplicated between `impersonate.ts`/`http.ts`) is tracked as #355, also not addressed here.

## Testing

- `pnpm --filter @lurkloot/core --filter @lurkloot/cli typecheck`
- `pnpm --filter @lurkloot/cli test` — 149 passed, including 3 new cases (WAF-blocked JSON envelope, non-WAF 403 → `authentication_rejected`, non-JSON HTML challenge page → `KickWafBlockedError`)
- `pnpm -r typecheck` and `pnpm test` (full workspace) — all green

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of Kick HTTP errors, including authentication failures and access-blocking security challenges.
  * Added clearer detection of Cloudflare-style challenge responses.
  * Prevented regular “Invalid session” responses from being misidentified as security blocks.
* **Tests**
  * Expanded coverage for blocked-access and authentication error scenarios.

### DIFF
--- a/packages/cli/src/transport/cycle.ts
+++ b/packages/cli/src/transport/cycle.ts
@@ -1,5 +1,6 @@
 import initCycleTLS, { type CycleTLSClient, type CycleTLSWebSocketResponse } from "cycletls";
-import { KickWafBlockedError } from "@lurkloot/core/tabs";
+import { KickWafBlockedError, safeKickFailure } from "@lurkloot/core/tabs";
+import { SafeFetchError } from "@lurkloot/core/fetchError";
 import type { PageFetcher } from "@lurkloot/core/adapter";
 import type { WebSocketFactory, WebSocketLike } from "@lurkloot/core/kick/watch";
 import type { PlatformCredentials } from "../authStore";
@@ -62,11 +63,24 @@ export function createCycleKickFetcher(cycleTLS: CycleTLSClient, creds: Platform
         body: typeof init?.body === "string" ? init.body : undefined,
       }, method);
 
-      if (response.status === 403) {
-        throw new KickWafBlockedError(`HTTP 403 from ${new URL(url).host} (Cloudflare blocked the impersonated request)`);
-      }
       if (response.status >= 400) {
-        throw new Error(`HTTP ${response.status} from ${new URL(url).host}`);
+        // Mirrors fetchKickInBackgroundWith's classification so checkAuthHealth sees the
+        // same failure kinds (WAF challenge vs. rejected credentials) over this transport
+        // as it does over the extension's page-context fetcher. safeKickFailure only
+        // recognizes Cloudflare's block by matching Kick's JSON error envelope; against
+        // this impersonated session a 401/403 is just as likely to come back as an HTML
+        // challenge page (see tabs.ts's own "likely a challenge page" handling), which
+        // would otherwise silently fall through to "authentication_rejected" and could
+        // suspend farming on what was actually a WAF block, not a bad session token.
+        const text = typeof response.data === "string" ? response.data : JSON.stringify(response.data ?? "");
+        const isJsonBody = (typeof response.data === "object" && response.data !== null) || safeJsonParse(text) !== undefined;
+        if (!isJsonBody && (response.status === 401 || response.status === 403)) {
+          throw new KickWafBlockedError(`HTTP ${response.status} from ${new URL(url).host} — non-JSON body (likely a Cloudflare challenge page)`);
+        }
+        const failure = safeKickFailure(response.status, text);
+        throw failure.kind === "security_policy_blocked"
+          ? new KickWafBlockedError(failure)
+          : new SafeFetchError(failure);
       }
 
       const data = response.data;
--- a/packages/cli/tests/impersonate.test.ts
+++ b/packages/cli/tests/impersonate.test.ts
@@ -1,5 +1,6 @@
 import { afterEach, describe, expect, it, vi } from "vitest";
 import { KickWafBlockedError } from "@lurkloot/core/tabs";
+import { isSafeFetchError } from "@lurkloot/core/fetchError";
 import { createImpersonateTransport } from "../src/transport/impersonate";
 import { createTvLinkAuthenticator } from "../src/transport/cycle";
 import { CHROME_JA3 } from "../src/transport/common";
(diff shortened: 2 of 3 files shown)


jamezrin:
Reviewed. **Correct and ready — the only thing holding it is the missing approval**, not CI. All checks pass (`verify`, both Docker arches, extension build/finalize, CodeRabbit); `mergeStateStatus` is `BLOCKED` purely because `reviewDecision` is empty and branch protection wants a review.

The fix is the right shape. Exporting `safeKickFailure` mirrors the existing `KickWafBlockedError` export with the same justification, and the added comment says why. The non-JSON guard is a genuine addition over what #354 described rather than a copy of it — `safeKickFailure` only recognises a WAF block by regex-matching Kick's JSON error envelope, so a Cloudflare HTML challenge page would otherwise fall through to `authentication_rejected` and could suspend farming on what was actually a WAF block. Good catch, and the test for it is the valuable one of the three.

Also worth noting the scope discipline: correctly dropping the #353 half after #351 fixed it independently, and leaving the `impersonate.ts`/`http.ts` duplication to #355 rather than widening the diff. `packages/core` gains no browser dependency, so `coreBoundary.test.ts` stays satisfied.

## One edge case, non-blocking

A 403 with a **null or absent body** is classified as `authentication_rejected` rather than reaching the challenge-page guard — the inverse of the misclassification this PR fixes.

```ts
const text = typeof response.data === "string" ? response.data : JSON.stringify(response.data ?? "");
const isJsonBody = (typeof response.data === "object" && response.data !== null) || safeJsonParse(text) !== undefined;
```

When `response.data` is `undefined` or `null`, `?? ""` makes `text` the string `'""'` — which is *valid JSON* — so `safeJsonParse` returns `""`, `isJsonBody` becomes `true`, the guard is skipped, and `safeKickFailure(403, '""')` sees an empty body, finds no reason, and returns `authentication_rejected`. Traced through every case:

| `response.data` | `text` | `isJsonBody` | classified as |
|---|---|---|---|
| `{ error: "Request blocked by security policy." }` | `{"error":"Request blocked…"}` | true | WAF ✅ |
| `{ error: "Invalid session" }` | `{"error":"Invalid session"}` | true | `authentication_rejected` ✅ |
| `"<html>…Cloudflare…</html>"` | the HTML | false | WAF ✅ |
| `""` | `""` | false | WAF ✅ |
| `undefined` / `null` | `'""'` | **true** | **`authentication_rejected`** ❌ |

The empty-*string* case works; only the null/absent case falls through, because `?? ""` routes it into `JSON.stringify` instead of treating it as an empty body. A body-less 403 is a plausible shape for a bare Cloudflare block, and this is the one input where the wrong answer has the consequence the PR exists to prevent — reporting `invalid_credentials` for a WAF block would tell the user to re-authenticate a session that is fine.

One line:

```ts
const text = typeof response.data === "string"
  ? response.data
  : response.data == null ? "" : JSON.stringify(response.data);
```

That collapses null/undefined onto the already-correct empty-string path and leaves every other row unchanged. Worth a fourth test case alongside the existing three.

Your call whether to fold that in before merging or take it as a follow-up — the current behaviour is still strictly better than the `every 403 is a WAF block` it replaces, so it isn't a reason to hold the PR.

PR #435, opened 2026-08-27
fix(popup): stop the popup's own resize cancelling drag reorder

### LINKED ISSUE #431: [Bug]: Unable to drag and reorder Drops or Idle Watchlist
### What happened?

Since I believe the last update, the ability to drag and reorder the priority for drops or idle watchlist items does not work. The cursor changes to indicate that it can be 'grabbed' but it does not grab or allow reordering anymore.

### Steps to reproduce

1. Open Lurkloot
2. Enable Twitch
3. Attempt to click and drag the 6 dot icon up or down.

### Where are you running Lurkloot?

Chrome or Chromium extension

### Lurkloot version

1.12.0

### Browser and operating system

Brave on Windows 11

### Affected streaming platform

Not platform-specific

### Watch mode

Not applicable

### Relevant logs

```shell

```

### Additional context

_No response_

### Before submitting

- [x] I searched the existing issues for this problem.
- [x] I removed passwords, cookies, tokens, authorization headers, and other credentials from this report.

jamezrin:
## Summary

Dragging the grip handle to reorder campaigns or Idle Watchlist channels does nothing in the extension popup. Fixes #431.

A browser action popup sizes its window to its content, so it re-measures — and fires `resize` — continuously, at an **unchanged** size. dnd-kit cancels any in-flight drag on `resize`, in **both** `PointerSensor` (via `AbstractPointerSensor`) and `KeyboardSensor`. So starting a drag fires the very event that kills it, and the reorder never completes.

Instrumented in a real popup, every attempt looked like this — note the drag *does* activate (188px of movement, well past the 5px constraint), is only ever "moved over" itself, and ends in `onDragCancel`:

```
10629ms pointerdown svg ON HANDLE
10816ms DND Draggable item ... was moved over droppable area ...   <- activated
10818ms resize 400x600 | doc 400x600                               <- 2ms later, SAME size
10824ms DND Dragging was cancelled.
```

`resize` also streams while the popup is idle, so even a drag that survived its own start gets killed shortly after — which is why keyboard reordering (Space + arrows) is broken too.

### This is not a regression

Locally built and manually tested: **v1.12.0, v1.11.0 and v1.10.1 all fail**, spanning two unrelated popup layouts (`SubTabs` vs the collapsible sections). Reorder itself is unchanged since the popup-ui package was extracted, and `@dnd-kit/*` has not moved. The cancel-on-resize behaviour is deliberate dnd-kit design since [PR #376, July 2021](https://github.com/clauderic/dnd-kit/commit/aede2cc42d488435cf65f19b63ba6bb7702b3fde), with no opt-out.

It went unnoticed because every place it was exercised lacks a popup auto-sizer:

| where | auto-sizes? | reorder |
|---|---|---|
| marketing site demo | no | works |
| `popup.html` opened as a tab | no | works |
| **real action popup** | **yes** | **fails** |

Reorder's only existing coverage is jsdom (`dropsView.test.tsx`), which calls `onReorder` directly and never exercises a popup window.

## The fix

`suppressPhantomResize()` swallows resize events that did not change the window size, registered at popup bootstrap. `resize` targets the window, so window listeners run in **registration order** — registering before React mounts puts this ahead of the listeners dnd-kit adds when a sensor starts.

A genuine size change passes through untouched, so dnd-kit's stale-measurement guard still works. This deliberately avoids reaching into dnd-kit's private `windowListeners`/`handleCancel` internals (an earlier prototype did, and would break silently on a rename); it uses only public DOM API, and fixes both sensors at once.

Its one cost: it swallows no-op resizes for every listener in the popup, not just dnd-kit's. That seems fair — nothing changed, so any correct listener is a no-op — but it is a global interception and is commented as such.

## Testing

- `pnpm test` — green (1624 extension, 156 CLI), including 5 new tests in `popupResize.test.ts`
- `pnpm typecheck` — clean
- Built the extension and drove real CDP input against it; same matrix on stock vs patched:

| scenario | stock | patched |
|---|---|---|
| pointer, no event (control) | reorders | reorders |
| **pointer, spurious resize (same size)** | **CANCELLED** | **reorders** |
| pointer, real resize (size changed) | CANCELLED | CANCELLED |
| pointer, visibilitychange | CANCELLED | CANCELLED |
| **keyboard (Space/Arrow), spurious resize** | **CANCELLED** | **reorders** |

Only the two intended rows flip.

## Verification

- [x] **Verified by hand in a real popup** — reordering works. This was the one thing automation could not confirm: a headless action popup is fixed-size (measured 415x510, zero resizes on content change) and cannot reproduce the real phantom resize stream.
- [x] The guard compares `innerWidth`/`innerHeight`. Since suppression demonstrably fires in a real popup, `inner` is confirmed to be what the phantom resizes leave unchanged — had `outerWidth`/`outerHeight` been the moving value, the guard would never have matched and behaviour would be unchanged.

No permission or manifest changes.

### DIFF
--- a/packages/extension/entrypoints/popup/main.tsx
+++ b/packages/extension/entrypoints/popup/main.tsx
@@ -1,6 +1,11 @@
 import { createRoot } from "react-dom/client";
 import "@lurkloot/popup-ui/styles.css";
 import { PopupApp } from "./app";
+import { suppressPhantomResize } from "../../src/core/popupResize";
+
+// Must run before React mounts: dnd-kit registers its own `resize` listener
+// when a drag starts, and window listeners fire in registration order.
+suppressPhantomResize(window);
 
 // Thin extension bootstrap. All popup UI lives in ./app so it can also be
 // imported and rendered standalone (with mock data) by the marketing landing
--- a/packages/extension/src/core/popupResize.ts
+++ b/packages/extension/src/core/popupResize.ts
@@ -1,0 +1,40 @@
+/** A window-like target, narrowed so tests can drive this with a stub. */
+export interface ResizeTarget {
+  innerWidth: number;
+  innerHeight: number;
+  addEventListener(type: "resize", handler: (event: Event) => void, options?: boolean | AddEventListenerOptions): void;
+  removeEventListener(type: "resize", handler: (event: Event) => void, options?: boolean | AddEventListenerOptions): void;
+}
+
+/**
+ * A browser action popup sizes its own window to its content, so it re-measures
+ * — and fires `resize` — continuously, at an unchanged size. dnd-kit cancels any
+ * in-flight drag on `resize` (both `PointerSensor`, via `AbstractPointerSensor`,
+ * and `KeyboardSensor`), so starting a drag fires the very event that kills it
+ * and campaign/watchlist reordering never completes. See issue #431: the same
+ * popup UI reorders fine on the marketing site and in `popup.html` opened as a
+ * tab, because neither of those is an auto-sizing popup window.
+ *
+ * Swallow the phantom events before anything else sees them. `resize` targets
+ * the window, so window listeners run in registration order — calling this at
+ * bootstrap puts it ahead of the listeners dnd-kit adds when a sensor starts.
+ * A genuine size change is passed through untouched, so dnd-kit's
+ * stale-measurement guard still works.
+ *
+ * Returns a teardown function.
+ */
+export function suppressPhantomResize(target: ResizeTarget): () => void {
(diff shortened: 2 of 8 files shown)


jamezrin:
Superseded by #436, which migrates to `@dnd-kit/react`.

The workaround here suppressed the popup's phantom `resize` events so dnd-kit would stop cancelling the drag. `@dnd-kit/react` re-measures on resize instead of cancelling, so the root cause is gone and no workaround is needed — verified by building the migrated extension with no suppression of any kind and confirming reorder completes for both pointer and keyboard.

The native-drag fix from this PR (links/images inside sortable rows starting an HTML5 drag) is carried over to #436.

Issue #472, opened 2026-09-07
Release pipeline: improvements to bring the implementation back in line with its invariants


jamezrin:
Tracking issue for improvements found while reviewing the `preview` / `production` release pipeline (all nine workflows, `RELEASING.md`, `scripts/release/*`). The pipeline works; every item below is a place where the implementation is behind one of its own stated invariants, or where a failure mode has no written escape. Ordered by payoff.

- [ ] 1. Promote the Docker candidate by digest instead of rebuilding
- [ ] 2. Move the `develop` verification off the release critical path
- [ ] 3. Add build provenance and SHA-pin actions
- [ ] 4. Document (or automate) recovery from an abandoned release
- [ ] 5. Fix two `RELEASING.md` statements that no longer match the code

---

## 1. Promote the Docker candidate by digest instead of rebuilding

**Where:** `.github/workflows/release.yml:90` (`docker` job), publish step at `release.yml:203`.

**Problem:** The pipeline's stated invariant is that the shipped bits are the reviewed bits — the extension honours it by downloading the candidate assets back instead of rebuilding. The `docker` job does the opposite: it rebuilds OCI archives from the merged ref rather than retagging the already-verified `candidate-X.Y.Z`. Same source, different build. The publish step's own comment concedes the digests are not guaranteed to match, and works around it by leaving an existing `X.Y.Z` in place.

**To fix:** resolve the `candidate-X.Y.Z` manifest digest and promote *that* to `X.Y.Z` / `X.Y` / `X` / `latest` inside the approved job. The rebuild path can stay as an explicit recovery fallback, but it should not be the normal route. Once this lands, the "shipped bits are the reviewed bits" comment in `release.yml` applies to both artifacts and the workaround comment can go.

## 2. Move the `develop` verification off the release critical path

**Where:** `.github/workflows/release.yml:255-261`, `scripts/release/sync.mjs:25` (`defaultVerify`, called at `sync.mjs:83`).

**Problem:** `sync-branches` runs `pnpm verify` on the merged `develop` tree as the final step of the approved production job. Release completion is therefore coupled to `develop`'s health at publication time, on a tree that is not the one being released. If `develop` has advanced with something that fails verify, the release wedges at the last step, and each rerun repeats the entire publish before failing in the same place.

**To fix:** split the sync into its own job (or workflow) that `publish` triggers but does not block on, so a develop-side failure leaves a completed release plus one retryable job.

[...]

PR #478, opened 2026-09-07
ci(release): move the develop sync off the release critical path

jamezrin:
Item 2 of #472.

## Summary

`sync-branches` ran `pnpm verify` on the merged `develop` tree as the final step of the approved `publish` job. Release completion was therefore coupled to `develop`'s health at publication time, on a tree that is not the one being released: if `develop` had advanced with something that fails verify, the release wedged at its last step, and each rerun repeated the entire publish before failing in the same place.

- Synchronization is now its own `sync` job, needing `publish`. A develop-side failure leaves a completed release plus one retryable job.
- Its checkout is clean, so the blanket `git restore --staged --worktree -- .` is gone rather than narrowed. That line was correct today but silently discarded anything else an earlier step left behind; a fresh checkout removes the need entirely (the "while in there" note on the issue).
- `publish` keeps its fail-fast `sync-token` step, so missing App configuration still fails before anything is published. The `sync` job mints its own token.
- `pnpm` setup moved out of `publish`, which no longer runs pnpm at all.

## The one judgment call — say the word and I'll flip it

**The `sync` job needs the App credentials, which live on `production`, so it requests a second approval.** Environment protection is evaluated per job; there is no way to split the job and keep exactly one approval while the credentials stay on the protected environment. I picked the option that preserves both the credential boundary and the pre-push verification, and updated `RELEASING.md` / `AGENTS.md` accordingly — the cost is one extra click per release.

The alternatives, if you would rather keep the single approval:

1. Move `RELEASE_SYNC_APP_ID` / `RELEASE_SYNC_APP_PRIVATE_KEY` to repository secrets. The `sync` job then needs no environment: one approval, one retryable job — but the sync credential leaves the protected boundary. One-line change here.
2. Keep sync inside `publish` and drop `defaultVerify` from the pre-push path. One approval, fully decoupled — but nothing verifies `develop` before the push, and `pr-validation.yml` is PR-only, so a broken tree is not caught until the next PR (which is strict-up-to-date gated, so it does get caught, later and by someone else).

## Testing

Workflow and docs only; `sync.mjs` is unchanged, so its unit tests still cover the behavior as-is. `release.yml` re-parses and the job graph is `resolve → docker/site-build → publish → sync`.

https://claude.ai/code/session_01ASCiGmqtAcEfdebkmh1qEf

## Summary by CodeRabbit

- **Release Process**
  - Release publishing and post-release synchronization are now handled as separate, independently approved stages.
  - Post-release synchronization merges the released changes into the development branch after verification.
  - Recovery guidance now includes steps for handling synchronization failures.

- **Documentation**
  - Updated release procedures and contributor guidance to reflect the revised publishing and synchronization workflow.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -146,9 +146,6 @@ jobs:
           RELEASE_SYNC_APP_ID: ${{ secrets.RELEASE_SYNC_APP_ID }}
           RELEASE_SYNC_APP_PRIVATE_KEY: ${{ secrets.RELEASE_SYNC_APP_PRIVATE_KEY }}
         run: node scripts/release/cli.mjs app-token
-      - uses: pnpm/action-setup@v6
-        with:
-          version: 11.26.0
       - uses: actions/download-artifact@v8
         with:
           pattern: stable-docker-${{ github.run_id }}-${{ github.run_attempt }}-oci-*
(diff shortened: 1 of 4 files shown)


jamezrin:
One extra data point found while re-reading `.github/actions/deploy-site/action.yml`: since the 1.6.0 fix, wrangler installs into a scratch directory outside the repository and the action asserts the tree is clean afterwards. So the blanket `git restore` this PR removes was already guarding against something that can no longer happen — the clean checkout makes it doubly unnecessary. Publication also no longer runs pnpm at all, which is why the pnpm setup moved to the `sync` job.

https://claude.ai/code/session_01ASCiGmqtAcEfdebkmh1qEf

Issue #569, opened 2026-09-19
fix(popup): consolidate finished campaigns in a Completed section


jamezrin:
## What happened?

Campaign rows can show several contradictory signals after every reward has already been completed:

- **100%** progress in the row headline
- a **Later** pill whose tooltip says the campaign will be farmed later
- a **Finished** lifecycle pill
- an amber warning whose tooltip explains why the campaign is not being watched

This makes a completed campaign look queued, finished, and blocked at the same time. The same noise repeats for every completed campaign and takes space away from campaigns that still require attention.

The underlying cause is that these signals are derived independently. **Later** is based on the campaign's position relative to the currently farmed campaign, while **Finished** and the rejection warning come from lifecycle and farmability evaluation.

## Steps to reproduce

1. Open the Lurkloot browser-extension popup.
2. Enable Twitch or Kick and populate the Drops list with multiple campaigns.
3. Complete all rewards in one or more campaigns while another campaign remains active and is being farmed.
4. Keep finished campaigns visible in the Drops list.
5. Observe a finished campaign above the currently farmed campaign.

## Current result

The finished row can simultaneously render `100%`, `Later`, `Finished`, and a farming-rejection warning.

## Expected result

Finished campaigns should have one clear terminal presentation and should not compete visually with campaigns that can still be farmed.

## Agreed design

- Split campaigns into an active priority list and a **Completed** subsection.
- Place the Completed subsection below active campaigns.
- Collapse Completed by default and show its campaign count in the disclosure header.
- When expanded, show **Finished** as the row's single headline state with a check icon.
- Do not show the following on a finished row:
  - `100%` headline progress
  - `Later`
  - lifecycle status pill duplicating `Finished`
  - farming-rejection warning or expanded rejection explanation
  - priority rank or drag handle
  - edge progress bar
- Keep reward details available by expanding the finished campaign row.
- Preserve the stored positions of finished campaigns when active campaigns are reordered, so their relative priority is retained if campaign state changes.
- Keep finished campaigns discoverable through Drops search.
- If another popup action focuses a finished campaign, automatically open the Drops section, open Completed, expand the campaign, and scroll it into view.

## Important edge case

A watch reward at **100%** that is still claimable or otherwise unclaimed is not a finished campaign. It must remain in the active list and retain its progress presentation.

## Acceptance criteria

- [ ] Completed is collapsed by default and displays the correct count.
- [ ] Active campaigns remain visible and keep contiguous visible ranks.
[...]

Issue #567, opened 2026-09-19
test(twitch-extensions): live acceptance for tabless NoPixelV and Fortnite farming


jamezrin:
Parent: #506. Follows the implementation merged in #541 (closes #507–#511 as implemented).

## Goal

Run the remaining live acceptance checks for tabless Twitch Extension reward farming in an installed browser with a normal logged-in Twitch session and **no Twitch channel tabs**. Deterministic tests, mocks and anonymous public reads do not satisfy these checks. Requirement-by-requirement evidence is in [`docs/twitch-extensions/acceptance.md`](https://github.com/jamezrin/lurkloot/blob/develop/docs/twitch-extensions/acceptance.md).

Already confirmed live: linked, authenticated background authorization for both providers, and a server-confirmed NoPixelV giveaway entry on ssaab.

## Sessions and lifecycle (from #507)

- [ ] Live disable, grant revocation, logout and restart tear down provider sessions with no stale reactivation, on Chromium (116+) and Firefox.

## Supplemental watch lane (from #508)

- [ ] A provider discovers and watches a channel with no active drop campaign and no Twitch tab.
- [ ] Completion, revocation and manual pause release the lane and ordinary drop farming resumes.

## NoPixelV (from #509)

- [ ] Daily watchtime increases on a discovered eligible channel in a tabless run (60/60 alone does not prove growth).
- [ ] Pack delivery is observed, and opted-in automatic opening produces confirmed `pack_opened` activity.
- [ ] Completion, disable and grant-revocation behave correctly live.

## Fortnite session and rewards (from #510)

- [ ] Authenticate and join with a linked viewer identity in a logged-in background run without Twitch tabs.
- [ ] Confirm the private participant/channel schemas, collection and reward state, and Epic account-link outcomes.
- [ ] Verify live reconnect, expiry and authority teardown.

## Fortnite captures and takeovers (from #511)

- [ ] Tabless captures increment the participant collection and generate confirmed activity.
- [ ] Server-earned phase rewards are observed rather than inferred from collection size.
- [ ] An opted-in eligible takeover is confirmed as owned by the viewer; closed-window and cooldown behavior checked.
- [ ] Live restart/reconnect and lane release.

## Notes

No credentials, JWTs or network dumps are needed in any report. Once everything here is checked, close this issue and the epic after a final pre-release review.

Issue #566, opened 2026-09-19
Replace generic AI-default color branding with a distinct Lurkloot identity


jamezrin:
## Problem

Lurkloot's visual identity is built from generic defaults, the kind an AI design pass reaches for first. Other drop farmers already use the same choices, so Lurkloot looks like one more of them:

- https://dropilot.app/: its landing CSS uses Kick green `#53fc18` heavily, the same hue as our `--lime`.
- https://aalejandrofer.github.io/GrubDrops/: it uses `#53fc18` plus Bricolage Grotesque, the same display font as ours.

Right now the palette isn't ours. It's the two platforms' own brand colors plus a gradient between them:

| Token | Value | Origin |
| --- | --- | --- |
| `--purple` / popup `--accent` (Twitch) | `#9147ff` | Twitch brand purple |
| `--violet` / `--accent-strong` | `#7c3aed` | Tailwind `violet-600` |
| `--lime` / popup `--accent` (Kick) | `#53fc18` | Kick brand green |
| `--signal` | purple → lime `oklch` gradient | blend of the two above |
| `--ink-*` / `--void` | near-black blue-violet | the usual "dark SaaS" background |
| Fonts | Bricolage Grotesque + Geist / Geist Mono | the usual AI-generated pairing |

This combination (dark blue-black background, purple→neon gradient, glow shadows, Bricolage + Geist) is recognizably AI-generated. It also borrows platform trademarks as our identity, when we only need them to tag which platform a row belongs to.

## Goal

Create a Lurkloot brand of our own, deliberately designed and not "sloppy", and apply it consistently across the site, the popup, the store assets and the icons.

## Scope

- **Brand palette**: choose a primary accent that isn't Twitch purple or Kick green, plus a neutral ramp that isn't the default blue-black `--ink` scale. Define light and dark variants and check contrast (WCAG AA for text).
- **Platform colors become secondary**: keep Twitch/Kick colors only where the UI has to tell the platforms apart (badges, platform tabs), not as the product's accent or hero gradient. This changes the popup's per-platform `--accent` switch in `packages/popup-ui/src/styles.css`.
- **Typography**: reconsider Bricolage Grotesque + Geist. Pick a pairing that isn't the current default, or justify keeping one of them.
- **Motifs**: reconsider the purple→lime `--signal` gradient and the glow shadows (`--glow-purple`, `--glow-lime`, `--accent-glow`).
[...]

Issue #561, opened 2026-09-18
[Feature]: List of features for QoL


MangoSenpai:
### What problem would this solve?

General ideas for QoL features.

Mainly things to customize the amount of drops you see in the extension.

### What would you like Lurkloot to do?

Hey Jamezrin - Awesome project!

I'd love some of these features in the future:

1. Button for removing/filtering games from the main drops list (you would have to remove them from the block list in settings later on, to revert). This should be any drop, including sub badge drops.

2. Completed drops go to a new tab, essentially removing itself from the main drops list.

3. Priority settings for specific games or list, so that you don't have to drag a new drop to the top - if it's in your top priority. Maybe a star-button, or a list in settings.

4. Sub badge separated tab, so it's easier to see what badge drops are available, if you collect them.

5. Setting for displaying remaining time on drops in collapsed view, maybe just an indicator like: 1D, 12H, 50M depending on the time remaining. To save space this could display only the biggest indicator remaining: 7D for 1 week remaining, 23H when there is less than 1 day remaining counting down.

Thanks for your time, and this project. It has really made it much easier for me to get some of these drops done, while I work or play games.

### Which part of Lurkloot does this affect?

Browser extension

### Before submitting

- [x] I searched the existing issues for a similar request.
- [x] This request does not involve storing credentials, exporting cookies, or bypassing platform detection.

jamezrin:
Hello @MangoSenpai thanks for words! 
I'm considering your suggestions for a UI redesign. It will cover multiple of your suggestions by itself + some new features that really make the UI beg for the new design. It will probably come in v2 which I aim to get out within the next month. I will update this issue with news. 
Thanks again for your message and I hope you keep enjoying the extension!

PR #571, opened 2026-09-20
feat(popup): rebuild the popup as a sidebar workspace with one campaign ranking

### LINKED ISSUE #562: feat(popup): name the stream being watched manually
## Summary

When **Pause when watching manually** is enabled and LurkLoot detects visible playback in a user-owned Twitch or Kick tab, the popup now explains the cause (#406) but never says *which* stream caused it:

> Manual watching · Farming is paused while you're watching a stream yourself. Pause it, close it, or switch away from that tab to resume automatically on the next check.

With several Twitch/Kick tabs open — a background tab that quietly resumed playback, a second monitor, a stream left running after a channel switch — "that tab" is not actionable. The popup should name the channel it detected, so the instruction points at something the user can find.

The channel is **already captured**: `recordManualWatchTelemetry` stores `channel` on `ManualWatchState` (`packages/core/src/background/controller.ts`), the full `SchedulerState` reaches the popup through `RuntimeSnapshot`, and `manualWatch[platform]` is the selected record. This is a presentation change only — no new data collection, no new fetch, no permission or `host_permissions` change, and nothing leaves the browser.

## Scope

- [ ] Render the manually-watched channel in `AutomationStatusLine` (`packages/popup-ui/src/automation.tsx`) for the manual-watch paused state. `detailKey` is a bare key rendered as `t(presentation.detailKey)` with no arguments, so the name cannot flow through the existing detail string — give manual watch its own render branch, mirroring the `presentation.state === "running" && farmingChannel` branch, so the channel can be an `<a href={channel.url}>` with `LINK_CLASS` and the `max-w-[7.5rem] truncate` treatment.
- [ ] Feed the channel into `automationPresentation()` (`packages/popup-ui/src/automationStatus.ts`) as a new input alongside the existing `manualClosePaused`, wired in `Popup.tsx` from `snapshot.state.manualWatch?.[id]`. The popup is currently the only consumer, so the signature change has no other host to update.
- [ ] Display `displayName ?? username`. `twitchChannelFromUrl`/`kickChannelFromUrl` derive the record from the tab URL and populate only `platform`, `username` and `url` — no `displayName`, `viewerCount` or `title`. Do **not** add a lookup just to recover display casing, and do not copy the viewer/campaign pills from the running branch.
- [ ] Keep the existing detail copy when no channel is known (a record can exist without one), and keep the line's fixed-height behaviour: the five-second snapshot poll must not be able to resize the header.
- [ ] Gate on the *record*, not on a second read of `pauseOnManualWatch`. The setting is already structurally enforced upstream — `recordManualWatchTelemetry` only marks a record `active` when `settings.pauseOnManualWatch`, and `updateManualWatchTab(..., enabled = false)` deletes the platform's records outright — so re-reading the setting in the UI only creates a second source of truth that can drift.
- [ ] Anchor the name on the state that already triggers the presentation (`session.reasonCode === "manual_watch"`), not on a `manualWatch` record merely existing, so a record stale past `MANUAL_WATCH_TTL_MS` can never name a streamer while farming is running. `hasRecentManualWatch` is controller-local and unexported — do not reimplement the TTL check inside the popup.
- [ ] Show only the selected record (`manualWatch[platform]`). `manualWatchTabs` may hold several tabs; this is not a list.
- [ ] Add the new copy to all catalogs under `packages/locales/messages/` (including RTL `ar`). This is host UI copy localized normally — not a diagnostic, so no `diagnostic*` key.
- [ ] Cover it in `packages/extension/tests/automationStatus.test.ts` (pure presentation) and `automationView.test.tsx`: channel present, channel absent, stale record, and **Pause when watching manually** turned off (no name, no manual-watch state).

## Out of scope

- Naming the channel in persisted activity events or notification copy. Deliberately excluded to keep this to the popup status line.
- Any change to when farming pauses or resumes. Detection behaviour is untouched.

## Reproduction

1. Enable Twitch or Kick and leave **Pause when watching manually** enabled.
2. Start farming, then open a stream on that platform in another tab and watch it.
3. Open the LurkLoot popup: the manual-watch pause is explained, but the streamer is not named.

Related: #406 added the pause explanation this builds on; #471 and #524 cover keeping claims running during manual-watch pauses.

### LINKED ISSUE #563: feat(extension): add the current channel to the idle watchlist from the in-page button
The in-page nav button (`packages/extension/src/core/inPagePanel.ts`) currently does exactly one thing: open the floating panel. When you are already standing on a channel page, the fastest way to farm that channel is still "open the panel → expand Idle Watchlist → press + → type the name you are literally looking at". Adding the channel you are watching should be one click from the page itself.

Turn the single nav button into a **split control**: the main area keeps its current one-click `Open Lurkloot` behaviour, and a caret opens a small menu whose items are contextual to the page.

Proposed menu:

| Item | Shown when |
| --- | --- |
| Open Lurkloot | always (redundant with the main area, but a menu with one contextual item reads oddly without it) |
| Add `<channel>` to the idle watchlist | on a channel page, channel not already listed |
| Remove `<channel>` from the idle watchlist | on a channel page, channel already listed |

One anchor, one nav footprint. `resolveAnchorFor` is already documented as the fragile part of this feature on both sites — Kick's nav class changed spelling between two page loads during development — so a second independent button next to the channel name would double that exposure for no gain.

Scope:

- [ ] Split the nav button into a main action plus a caret that opens a hand-written menu. Keyboard accessible (Enter/Space, Escape, arrow keys), closes on outside click and on Escape, and does not trap focus.
- [ ] Add a pure, exported `channelFromLocation(platform, url)` with reserved-path handling, tested in `packages/extension/tests/` the way `resolveAnchorFor` already is. Twitch: skip `/directory/*`, `/videos/*`, `/drops/*`, `/settings/*`, `/subscriptions`, `/wallet`, `/u/*`, `/popout/*`, `/moderator/*`, and accept `/<login>` plus its `/about`, `/schedule`, `/videos`, `/clips`, `/clip/*` subpaths. Kick: skip `/browse`, `/categories/*`, `/following`, `/dashboard/*`, `/search`, and accept `/<slug>` plus `/videos`, `/clips`, `/about`.
- [ ] Recompute the channel **when the menu opens**, not on navigation. Both sites are SPAs; the existing code re-anchors on mutation/resize/fullscreen and never cares which channel it is on. Lazy recompute sidesteps all pushState/popstate plumbing and cannot go stale.
- [ ] Hide the watchlist item entirely when the path is not a channel page, rather than showing it disabled.
- [ ] Write through the existing `{ type: "saveSettings", settingsPatch, tickAfterSave: true, tickAfterSavePlatforms: [platform] }` message (already in `CoreRuntimeMessage`, dispatched at `packages/extension/entrypoints/background.ts:339`). Never write `storage.local` directly — that skips `normalizeChannelList` and the tick.
- [ ] Reflect the result in the menu (item flips add ↔ remove) and keep the panel's watchlist view in sync, which it already is via the storage listener.
- [ ] Add tests for the menu open/close, the add and remove paths, the message shape, and the reserved-path table.

Constraints:

- [ ] **No new imports into `inPagePanel.ts`.** WXT builds content scripts as a single IIFE with `inlineDynamicImports`, so anything imported here ships on every `twitch.tv` and `kick.com` page load — rendering the popup in the content script cost 1.2 MB per platform, which is why this module is dependency-free and hand-styled. The menu is hand-written DOM with inline CSS; no `@lurkloot/popup-ui` primitives, no `lucide-react`, no Tailwind. The content-script bundle must stay single-digit KB, and the new-tab-per-page-load cost of the feature must be zero until the menu is opened.
- [ ] Menu strings use `browser.i18n.getMessage`, not literals. `wxt.config.ts` ships `default_locale: "en"` and materializes `_locales/*/messages.json` from `@lurkloot/locales`, and `browser.i18n` is a browser API with no bundle cost — the header comment's "cannot reach the locale catalogs" is about *importing* the package, not about `browser.i18n`. The existing `Open Lurkloot` / `Lurkloot` literals can move the same way in the same PR.
- [ ] The control stays subject to every existing gate: hidden when `showInPagePanel` is off, hidden inside a managed farming tab (`isManagedTab`), hidden in fullscreen.

Open questions:

1. **Watchlist cap.** `normalizeChannelList` (`packages/shared/src/settings.ts:371`) deduplicates and lowercases but does **not** cap the list; the `20` in `idleWatchlist.tsx` is display-only text on the section header. So the in-page path would bypass a limit the popup only appears to enforce. Either enforce a shared cap in `normalizeChannelList` (and make the popup's `+` respect it too), or accept no cap and drop the `/20` from the header. Pick one — the current state is a UI claim with nothing behind it.
2. **Concurrent edits.** `applySettingsPatch` replaces `idleWatchlistChannels` wholesale, so the menu must read a fresh snapshot, append/remove, and send the full array — last-write-wins against a popup open on the same watchlist at the same moment. That is acceptable; inventing merge semantics for a two-surface race that costs one re-add is not.
3. **Caret affordance width.** Both sites' navs are tight, and the button already carries icon + `Lurkloot` label. Worth checking on the live pages whether the caret should replace the text label at narrow widths.

Decided:

- Removing a channel while it is the active idle watch behaves exactly as the same edit made from the popup does — released on the next tick. No new behaviour.

Alternatives considered:

- A separate small button next to the channel name (as originally sketched). Rejected: a second anchor on a second, less stable part of each site's DOM, and it puts a Lurkloot control in the middle of the streamer's own header.
- Making the button menu-only, with `Open Lurkloot` demoted to a menu item. Rejected: opening the panel is the common action and should stay one click.

Non-goals: toggling `excludedChannels` / the block list from the page, reordering the watchlist from the page, and any in-page control on non-channel pages beyond `Open Lurkloot`.

Related: #561 collects other QoL requests; the floating panel itself landed in #537 and the surrounding 1.13.0 work.

### LINKED ISSUE #564: feat(popup): allow changing a campaign's rank while searching
## Problem

When the campaign list in the Drops panel is filtered with the search box, a campaign's priority rank can no longer be changed. The rank number still shows each campaign's real position in the full list, but it is plain text rather than the editable rank input, and there is no drag handle.

With many campaigns, searching is the fastest way to find the one you want. To move it up or down, you currently have to close the search, scroll the full list to find it again, and change the rank there.

## Current behaviour

In `packages/popup-ui/src/drops.tsx` (`DropsPanel`), the search branch renders a plain `CampaignCard` without `rankCount` or `onRankMove`, and outside the `DragDropProvider`. With no `onMove`, `RankInput` (`packages/popup-ui/src/primitives.tsx`) falls back to a read-only `<span>`.

## Proposal

Make the rank input editable while searching. Typing a new rank moves the campaign to that position **in the full (unfiltered) priority list**. It does not move it within the filtered results.

- Pass `rankCount={campaigns.length}` and an `onRankMove` that maps the card's full-list index (`priorityIndex`) to `moveCampaign(priorityIndex, toIndex)`.
- Keep the search query and results open after the move. The card stays in the filtered list, and its rank number updates to the new position.
- Drag-to-reorder can stay disabled while searching. Dragging inside a filtered subset is ambiguous (where do the hidden campaigns end up?), so the rank input is the clearer control there.

## Acceptance criteria

- [ ] While a search query is active, clicking a campaign's rank opens the editable rank input.
- [ ] Entering a rank moves that campaign to that position in the full priority order, and the change is saved like any other reorder (`onReorder`).
- [ ] The search stays open with the same query after the move.
- [ ] Unit tests cover reordering from a filtered view.

### LINKED ISSUE #565: fix(popup): “Later” status appears in Spanish with English selected
## What happened?

With Lurkloot's language set to English, the Drops list can display **“Más tarde”** on waiting campaign status chips while the surrounding popup text remains in English (for example, “Finished,” “Farming,” and “Idle Watchlist”).

**Expected:** The chips should say **“Later”** when English is selected.

## Reproduction

1. Use a browser whose extension UI messages resolve to Spanish.
2. Select English in Lurkloot's language setting.
3. Open the popup's Drops list with campaigns waiting behind the currently farmed campaign.
4. Observe **“Más tarde”** on the waiting chips while the surrounding text is English.

The user's browser and extension version have not been confirmed, but this locale combination follows deterministically from the current code.

## Root cause

`CampaignCard` in `packages/popup-ui/src/drops.tsx` renders the chip from `t("later")`. The English catalog correctly defines `later` as the literal string `later`, while the Spanish catalog defines it as `más tarde`.

In `packages/popup-ui/src/Popup.tsx`, `t()` calls `translateFromCatalogs()` and then treats **`message === key`** as a missing translation. For the valid English translation of `later`, that comparison is true, so `t()` calls `adapter.getMessage("later")`. The extension adapter uses `browser.i18n.getMessage()`, which returns the browser's Spanish message independently of Lurkloot's English override. Other labels such as `finishedPill` and `farmingLabel` do not equal their keys in English, so they stay English.

The English `live` message also equals its key and can cross the same fallback path. Presence of a catalog entry, rather than equality of its translated text to the key, needs to determine whether a translation is missing.

## Acceptance criteria

- [ ] With a Spanish browser UI and English selected in Lurkloot, waiting campaign chips render “Later” and no other English translation is replaced by a browser-locale message solely because its text equals its key.
- [ ] Other selected languages continue to use their own translations.
- [ ] Add a regression test for a valid translation whose text equals its key (including `later`).

## Screenshot

[User-provided screenshot](https://media.discordapp.net/attachments/1530382420472238272/1550650994994446548/image.png?ex=6aaf1bc6&is=6aadca46&hm=8ad0f756803ceb3444e6fa1b81ef7c3935d9ff828c2a2dda964ee651cf4954eb&=&format=webp&quality=lossless)

jamezrin:
## Summary

The popup was one 400 × 600 column. Chrome, platform picker, status line and
list toolbar stacked above a single scroll that carried the campaigns, the
Twitch extensions panel and the Idle Watchlist together, while campaign order
came from a hidden stack of four keys that the popup and the scheduler sorted
differently.

This rebuilds it as a **720 × 600 workspace**: a rail with the platform switch
and one entry per destination — Queue, Completed, Games, Idle watchlist,
Extensions (Twitch only), Activity, Settings — with the now-farming strip above
all of them. Both browsers cap a popup at 800 × 600, so height was already
spent; the unused width is what this takes.

Campaign order becomes **one function**, `rankCampaigns` in `@lurkloot/shared`,
shared by the scheduler, the keep-watching comparison and the Queue:

1. pinned campaigns, in pin order (`campaignPins`, sparse — dragging one card
   pins that card, not the whole list);
2. campaigns of favourite games, in the order they were starred;
3. everything else by one live strategy, then name, then id.

Eligibility is always decided before ranking, so a pin can never rescue a
campaign an exclusion, a block or a class filter refused. `campaignSection` puts
every campaign in exactly one list — queue, skipped, upcoming, completed,
expired — so the six `dropsListFilter` toggles retire with the campaigns they
used to hide.

## Issues

- Closes #565 — `t()` treated a catalog message equal to its key as missing, so
  English `later`/`live` fell through to `browser.i18n` and answered in the
  browser's locale.
- Closes #564 — a search result now carries the rank editor; the rank shown is
  its place in the full queue and typing there moves it among the pins without
  closing the query.
- Closes #562 — the manual-watch pause names the stream that caused it, as a
  link, gated on the session's own reason code.
- Closes #563 — the in-page nav button is a split control; its menu adds or
  removes the channel the page is about, through `saveSettings`.
- Refs #352 — implements approach C, extended with favourite games.
- Refs #561 — 1 (block a game), 2 (Completed tab), 3 (star a game), 4 (sub-badge
  facet); 5 was already covered by the compact time pills.
- Refs #566 — the structural half only, see below.
- **#569** was closed by #570 while this branch was in flight. Its agreed design
  was a collapsed subsection under the active list; this replaces that with the
  Completed destination #561.2 asked for, keeping every per-row rule from #569
  (one terminal state, no rank, drag handle, progress, `Later` or warning;
  rewards still expand; search and focus navigation still reach it).
- **#567 is untouched.** It is explicitly live acceptance — "deterministic
  tests, mocks and anonymous public reads do not satisfy these checks" — so it
  still needs a run in an installed browser. The deterministic supplemental-lane
  coverage it sits on top of is unchanged.

## Decisions worth reviewing

- **Not a breaking change.** Schema v7 migrates both hosts on load: the dense
  `campaignPriorities` map becomes ordered pins, `priority_list_only` becomes
  the `farmPinnedOnly` switch, `categoryMode: "exclude"` becomes `"all"` plus
  `blockedCategories`, and `dropsListFilter` is dropped. A hand-written CLI
  config keeps loading, with a deprecation diagnostic naming each moved key.
- **A favourite game never interrupts progress.** Only a pin counts as an
  explicit override for keep-watching; a star is a standing preference that
  re-fires as campaigns churn, and treating it as an override would abandon a
  healthy watch at 55/60 minutes whenever a starred game's campaign appeared.
- **Category order stops ranking.** The allowlist filters, `favouriteCategories`
  rank, `blockedCategories` deny in both modes.
- **#566, structural half only.** `--accent` no longer resolves to Twitch purple
  or Kick green per platform; it is one product token on `:root`, with platform
  hues confined to the rail's switch and badges. The *value* is deliberately
  left close to the previous violet — choosing the palette is #566's call, and
  it is now a one-token swap. `docs/brand.md` records the rule, the candidates
  and what remains (icons, store assets, site tokens, AA contrast).
- **The watchlist cap is now real.** The popup printed `n/20` while nothing
  enforced it; `normalizeChannelList` caps the list, the popup's `+` disables at
  the limit, and the in-page menu stops offering an add that would be dropped.
- **The view swaps outright instead of cross-fading.** The rail makes navigation
  frequent, and an exit animation held the outgoing panel on screen every time.

## Testing

- `pnpm verify` — script tests, workspace typechecks, 2,339 extension tests, 198
  CLI tests, the Astro site build, and both Chromium MV3 and Firefox MV2 builds.
- New suites: `ranking`, `rankingSettings`, `popupShell`, `queueView`,
  `gamesView`, `channelFromLocation`, `inPageMenu`, `popupTranslator`, plus a
  v6-document migration through the real `loadSettings` path.
- Content script stays hand-written and dependency-free: `twitch.js` is 16.06 kB
  built, about 1 kB more than before for the menu.

## Screenshots

Popup UI changed throughout, so screenshots are needed before merge. Two
knock-on changes to capture with them: the site's live demo now scales the real
720px popup instead of framing it at 400, and `demo.ts` was updated for the new
model (Starfall Arena is a favourite, Kick blocks Just Chatting), which is what
the CWS screenshot variants render.

## Summary by CodeRabbit

* **New Features**
  * Introduced a redesigned 720×600 popup workspace with navigation rail, Queue, Completed, Games, and Watchlist views.
  * Added campaign pinning, favourite-category ranking, blocked categories, pinned-only farming, search, filters, and queue management.
  * Added finished and expired campaign tabs with dedicated status views.
  * Added in-page controls to open the popup and manage idle-watchlist channels.
  * Added clearer manual-watch status with linked channel details.
  * Added platform-specific Twitch and Kick accent colors.

* **Bug Fixes**
  * Improved migration of legacy campaign priorities, category exclusions, and display filters.


jamezrin:
## Visual pass

Rendered the real popup at 720 × 600 through the site's live demo (the same component the extension loads) and walked every destination. Screenshots: **https://claude.ai/artifact/8jSqRnB7eac1JtRRMkYXmn** — private, so they need sharing before anyone else can open them; the PNGs are also reproducible from `pnpm dev:site`.

Four things it caught that the 2,343 tests could not, fixed in a489b7d:

- three tips still described the model this replaces ("drag campaigns by the grip to set their order", category filters in platform settings, excluded campaigns restored from settings);
- a game with one live campaign read "1 campaigns live";
- the Idle Watchlist kept its collapse toggle inside its own destination, where nothing can fold away — a control that did nothing;
- the rail counted only finished campaigns while the view it opens holds finished *and* expired.

The demo data also gains one finished and one expired campaign, so Completed is a populated view in the site demo and in the store screenshots rather than an empty shell.

Still not covered visually: dark theme, Kick, and the Skipped/Upcoming groups — the demo data carries no skipped or upcoming campaigns.

